PR #9, opened 2016-08-23
second entry fix

apmccartney:
Corrected a bug in which the second section of a file, the second file of a material, and the second material of a tape would include the (respective) first within the buffer.


jlconlin:
Nice catch.

PR #10, opened 2016-11-03
Feature/directory structure

apmccartney:
it begins


jlconlin:
Don't ever do this again. It's painful.

PR #14, opened 2017-06-02
Feature/copyable tab1

apmccartney:
see `UnivariateTable.test.cpp` demonstrating copyability.

Added ranges v3 dependency in anticipation of simplifying the LIST and TAB1 interfaces


apmccartney:
closes #13

PR #27, opened 2017-08-29
Hotfix/const correct

### LINKED ISSUE #26: operator==() in EndfTk record implemetations should be const
The operator==() functions implemented for the list and tab1 are not const. These are the only ones that we actually looked at but the other ones should be checked as well (you never know).

apmccartney:
closes #26


jlconlin:
Why the change to the underlying map implementation?

apmccartney:
Performance

apmccartney:
Don't merge, tests to come

whaeck:
I was going to wait to ask you about the coverage results before merging it anyway :-)

PR #31, opened 2017-08-31
Adding some 'getters' to file::Type.

jlconlin:


src/ENDFtk/file.hpp
```diff
@@ -1,17 +1,40 @@
 namespace file{
 
-template< int MF >
+template< int fileNo >
```
whaeck: I'll be a purist here, but shouldn't an MF number be unsigned? The same would go for the MT number as well.

I would personally not have changed the name from MF to fileNo because MF is the ENDF qualified name of the file type.

jlconlin: You'll have to talk to @apmccartney about `unsigned int`s. He has a good reason for not liking them.

I used `fileNo` here because I wanted a function called `MF` which returns the file number. I couldn't name both of them `MF`.

src/ENDFtk/file/Type/src/sectionNumber.hpp
```diff
@@ -0,0 +1,20 @@
+const Section&
+sectionNumber( int sectionNo ) const {
+  try{
+
+    return this->sectionMap.at( sectionNo );
+
+  } catch( std::out_of_range& o ){
+    Log::error( "Requested section number (MT) does not"
+                " correspond to a stored section in the syntax tree" );
```
apmccartney: > in the syntax tree

copy/paste error?

src/ENDFtk/file.hpp
```diff
@@ -5,13 +5,32 @@ class Type {
 public:
   /* convenience typedefs */
   using Section = section::Type< MF >;
+  using iterator = typename std::vector< Section >::iterator;
+  using const_iterator = typename std::vector< Section >::const_iterator;
 
   /* fields */
   std::vector< Section > sectionVector;
   std::unordered_map< int, Section& > sectionMap;
```
apmccartney: mind if I update this to a hopscotch_map?

src/ENDFtk/file.hpp
apmccartney: In general, unsigned integers types are a pretty crummy way to document non-negativity requirement on integer arguments in languages which provide implicit conversion between signed and unsigned types (such as C/C++), [e.g.](https://wandbox.org/permlink/3Ll5v47RnIF0BoAX)

```cpp
#include <iostream>

int main(){
  if (-10 == 4294967286u ){   std::cout << "weird" << std::endl; }
}
```

Instead prefer assertions provided by either the assert macro in <cassert> (for runtime conditions) or static_assert (for compile time conditions).

That said, here we're defining a template specialization for a positive integer so it doesn't much matter. My preference would be a signed int.

src/ENDFtk/file/Type/test/Type.test.cpp
```diff
@@ -17,7 +17,21 @@ SCENARIO( "Testing generic case using file 3" ){
       long lineNumber = 0;
       
       StructureDivision division( begin, end, lineNumber);
-      REQUIRE_NOTHROW( file::Type<3>(division, begin, end, lineNumber ) );
+      file::Type< 3 > MF3(division, begin, end, lineNumber );
+
+      THEN( "the sections can be extracted" ){
+        REQUIRE( 3 == MF3.MF() );
+        REQUIRE( 3 == MF3.fileNumber() );
```
apmccartney: are these getters still provided?

src/ENDFtk/file.hpp
whaeck: OK. I understand.

src/ENDFtk/file.hpp
jlconlin: I tried to use a `hopscotch_map`, but new problems crept in. I didn't have the patience to track those down at the time.

src/ENDFtk/file.hpp
```diff
@@ -8,11 +8,16 @@ class Type {
   using iterator = typename std::vector< Section >::iterator;
   using const_iterator = typename std::vector< Section >::const_iterator;
 
+protected:
   /* fields */
   std::vector< Section > sectionVector;
-  std::unordered_map< int, Section& > sectionMap;
+  tsl::hopscotch_map< int, const Section* > sectionMap;
```
jlconlin: I tried doing this with a `Section&`, but got compiler errors. Is the `hopscotch_map` not compatible with references and thus has to have pointers?

PR #35, opened 2017-09-05
Feature/mf2 getters

jlconlin:
This is only for the resolved resonance range.


apmccartney:
I really don't like that an Lvalue provides getters to data logically belonging to its resonances. It's like going into my neighbors house to get a cup of sugar for a friend

apmccartney:
I also do not like the function to extract resonance energies. This is a processing function which isn't in the scope of a parsing library. Moreover, there isn't an application can for the function (the reconr starting grid is more complicated).

PR #37, opened 2017-09-17
Feature/print

apmccartney:
+ added `NC` getters to most types
+ added `print` function to most types
+ refactored the unresolved resonance classes
+ renamed the `UnivariateTabulation` to `TabulationRecord`
+ added `Lvalue` and `Jvalue`/`Resonance` classes to resonance types
+ resonance types provide `LRO` and `LRF` getters


src/ENDFtk/ControlRecord.hpp
```diff
@@ -54,4 +54,8 @@ class ControlRecord : protected record::Base< record::Real,
   operator!=( const ControlRecord& rhs ) const {
     return not( *this == rhs );
   }
+
+  int NC() const { return 1; }
```
jlconlin: What is `NC`?

src/ENDFtk/DirectoryRecord.hpp
```diff
@@ -48,5 +48,9 @@ class DirectoryRecord : protected record::Base< record::Integer< 33 >,
   bool
   operator!=( const DirectoryRecord& rhs ) const {
     return not( *this == rhs );
-  }  
+  }
+
+  int NC() const { return 1; }
```
jlconlin: What is `NC`?

src/ENDFtk/ControlRecord.hpp
jlconlin: Oh, I get it now. It's the number of lines in the ENDF file.

src/ENDFtk/resonanceParameters/resolved/BreitWigner/Lvalue.hpp
```diff
@@ -0,0 +1,27 @@
+class Lvalue {
```
jlconlin: I'm not real keen on calling this `Lvalue` instead of `LState`. To me an l-value is an integer quantity specifying an independent. Whereas the l-state is the l-value plus the rest of the information describing the physics.

src/ENDFtk/resonanceParameters/unresolved/EnergyDependent/Jvalue.hpp
```diff
@@ -0,0 +1,29 @@
+class Jvalue {
```
jlconlin: For the same reason I don't like the name `Lvalue`, I also don't like the name `Jvalue`.

jlconlin: Also, shouldn't the `Jvalue`/`JState` be a class inside `LState`?

apmccartney: <yoda> Do or do not; there is no should </yoda>

What motivates the preference?

src/ENDFtk/resonanceParameters/resolved/BreitWigner/Lvalue.hpp
apmccartney: I wasn't fond of either name. I checked openMC and DECE to see if there were other more respresentative or descriptive names for this aggregate, but they both seem to work in terms of the primitives without provides this abstraction.

That said, I favor `Lvalue` as it mirrors the naming in the resonance reconstruction repo and refers to a physical quantity domain experts will recognize, as opposed to `LState`, which is a term we invented and is less likely to have an immediate association

apmccartney:
`NC` comments resolved and deleted

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths/src/readLvalues.hpp
```diff
@@ -0,0 +1,17 @@
...
+  if ( NLS < 1 ){
+    Log::error( "NLS required to be positive" );
+    Log::error( "NLS value: {}", NLS );
+    throw std::exception();
+  }
+  
+  lValues.reserve( NLS );
+  while ( NLS-- ){ lValues.emplace_back( it, end, lineNumber, MAT, MF, MT ); }
+  return lValues;
+}
```
whaeck: This function is implemented multiple times, using a different Lvalue class. Maybe add this as a template function in a base class?

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths/test/EnergyDependentFissionWidths.test.cpp
```diff
@@ -63,8 +60,8 @@ SCENARIO( "Testing energy-dependent fission width unresolved resonances" ){
           base, begin, end, lineNumber, MAT, MF, MT );
 
       THEN( "the parameters can be verified" ){
-        REQUIRE( 5.7E3 == edfw.EL() );
-        REQUIRE( 4.0E4 == edfw.EH() );
+        REQUIRE( Approx(5.7E3) == edfw.EL() );
+        REQUIRE( Approx(4.0E4) == edfw.EH() );
```
whaeck: How come you did not have a roundoff issue before?

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent.hpp
```diff
@@ -8,15 +12,17 @@ class EnergyIndependent {
...
-
+  
 public:
+  #include "ENDFtk/resonanceParameters/unresolved/EnergyIndependent/src/ctor.hpp"
+  
   double EL() const { return this->el; }
   double EH() const { return this->eh; }
-  int NRO() const { return bool(this->ape); }
+  int NRO() const { return bool( this->ape ); }
   int NAPS() const { return this->naps; }
```
whaeck: As above.

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent.hpp
```diff
@@ -25,7 +31,21 @@ class EnergyIndependent {
   int LSSF() const { return this->lssf; }
   int NLS() const { return this->lists.size(); }
```
whaeck: As above: LRU, LRF, LSSF, NLS

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent.hpp
```diff
@@ -25,7 +31,21 @@ class EnergyIndependent {
...
-  auto LStates() const {
-    return ranges::make_iterator_range( this->lists.begin(), this->lists.end() );
+  auto lValues() const {
+    return
+      this->lists
+      | ranges::view::transform
+        ( []( const auto& list ){ return Lvalue( list ); } );
   }
+
+  int NC() const {
```
whaeck: Same as above.

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent/Lvalue.hpp
```diff
@@ -0,0 +1,17 @@
+class Lvalue {
+  const LIST& core;
+
+public:
+  Lvalue( const LIST& list ) : core( list ){}
+
+  double AWRI() const { return this->core.C1(); }
+  int L() const { return this->core.L1(); }
+  int NJS() const { return this->core.N2(); }
```
whaeck: Same as above.

jlconlin:
This is a good place for nitpicking. We can discuss why we want things one way vs another. And having it on GitHub makes it possible to keep a record of our decisions (if only we could search them.)

src/ENDFtk/ControlRecord/test/ControlRecord.test.cpp
```diff
@@ -7,7 +7,7 @@ using namespace njoy::ENDFtk;
 
 SCENARIO( "ControlRecord Tests", "[ENDFtk], [ControlRecord]" ){
   std::string line =
-    " 1.001000+3 9.991673-1          0          0          0          5 125 1451    1\n";
+    " 1.001000+3 9.991673-1          0          0          0          5 125 1451     \n";
```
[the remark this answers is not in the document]
apmccartney: This functionality would require a running total for `NC`, which could be supplied as an argument as MAT, MF, and MT are; easy to accomodate with an overload.

That said, until someone requests the functionality, I'm happy to defer that work, especially in light of the deadline at the end of the month.

src/ENDFtk/ListRecord/src/print.hpp
```diff
@@ -0,0 +1,49 @@
...
+                                  disco::ENDF, disco::ENDF,
+                                  disco::Integer< 4 >, disco::Integer< 2 >,
+                                  disco::Integer< 3 >, disco::ColumnPosition<5> >;
+
+    auto nFullLines = this->NPL() / 6;
+    auto partialLineEntries = this->NPL() - nFullLines * 6;
+    
+    auto entry = this->list().begin();
+
+    for ( ; nFullLines--; entry += 6 ){
```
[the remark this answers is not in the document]
apmccartney: For random access iterators, `entry += 6` and `std::advance(entry, 6)` are equivalent.
For less powerful iterator categories, there is no `+=` overload and `std::advance` increments `n` times.

src/ENDFtk/ControlRecord/test/ControlRecord.test.cpp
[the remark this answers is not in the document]
whaeck: I concur. No need to anticipate a user's request that adds little to the toolkit.

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths/test/EnergyDependentFissionWidths.test.cpp
apmccartney: On the platforms we test on, we don't have round-off then or now. That said, floating point equality is squirrelly. 

Just a good habit when working with floating point numbers.

apmccartney:
Given the suggested number of changes from `int` to `long`, it appears I'm pretty consistent in using `int` values.

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths/src/readLvalues.hpp
apmccartney: No need for a base class for what would be a static function, but point taken

src/ENDFtk/ControlRecord.hpp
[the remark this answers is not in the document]
apmccartney: The value-to-effort ratio is pretty low for these requests and it pessimizes storage on the client side when they use auto.

src/ENDFtk/ListRecord/src/print.hpp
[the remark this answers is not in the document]
apmccartney:
```cpp
nFullLines-- != 0
```

Eh, I think its noisy, but I'm probably in the minority so I'll give you this one

src/ENDFtk/ControlRecord.hpp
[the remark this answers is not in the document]
whaeck: I understand that. I would argue that if we use long (or any other type) as the type for integer values stored in an 11 character column of the ENDF file (meaning L1, L2, N1, N2, interpolation types and indices), any ENDF defined parameter should use the same type (even if it is not stored in a record). Hence the remark that NC, LRF, etc. should be long (again, if we use long as the storage type).
However, the largest integer value that I have ever seen in an ENDF evaluation is about 250000 (which is NC for ENDF/B-VII.1 U235 MF1 MT460). Some covariance sections also reach NC values of this order of magnitude. Larger integer values could occur in pointwise reconstructed cross section NP values (but I would be surprised that this would be larger than say 1000000 unless of course you use an outrageously small reconstruction tolerance). Taking this in consideration, I agree that int would be more than sufficient.

src/ENDFtk/resonanceParameters/src/read.hpp
```diff
@@ -1,8 +1,8 @@
-template< typename Iterator >
+template< typename Lvalue, typename Iterator >
 static std::vector< Lvalue >
-readLvalues( int NLS,
-             Iterator& it, const Iterator& end, long& lineNumber,
-             int MAT, int MF, int MT ) {
+read( int NLS,
```
jlconlin: Why did you change the name of the function?

src/ENDFtk/ControlRecord.hpp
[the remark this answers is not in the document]
apmccartney: > I  would argue that if we use long (or any other type) as the type for integer values stored in an 11 character column of the ENDF file (meaning L1, L2, N1, N2, interpolation types and indices), any ENDF defined parameter should use the same type (even if it is not stored in a record). 

why though

src/ENDFtk/resonanceParameters/src/read.hpp
apmccartney: I didn't. New function similar enough to the previous, git thought it was a moved file

src/ENDFtk/ControlRecord.hpp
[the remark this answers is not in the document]
apmccartney: despite being an 11 character field, `LRU`, for example, can only assume 3 values: 0, 1, 2.

src/ENDFtk/resonanceParameters/src/read.hpp
jlconlin: The function used to be called `readLvalues`, now it is called `read`—regardless of what git does.

src/ENDFtk/ControlRecord.hpp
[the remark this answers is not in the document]
whaeck: I know. It's just that I like uniformity and consistency over the entire toolkit :-)
If you do it this way for MF2 MT451, I will do it this way for everything I put in (starting with file 1).

PR #38, opened 2017-10-13
Feature/mf1

whaeck:
File 1 section 451 functionality has been added. All data (parameters, description and index) can be retrieved from the section and it can be printed. Only section 451 was implemented even though some residual coding remains for the other sections (constructors). This will be added later.

A template specialisation was made for file 1 because it can only contain a section 451, 452, 455, 456, 458 and 460. All other sections are undefined. A file 1 can be read and the 451 section can be extracted from it. File 1 can also be printed. I will add functionality later to mimic the interface of the generic file (yes, MT451() should go away at some point and will be replaced by MT( 451 ) instead, and iterators should be added as well).

The template file::Type< FileNumber > was modified to allow for generic printing. This assumes that print( iterator, MAT, MF ) is implemented on the section. As the section knows which MT it is, the MT number is not present in the function call. This was tested using file 3.

An open question: for the moment, printing leaves the trailing 5 spaces where the sequence number used to be. Reading an ENDF record/section/... does not require these trailing spaces to be present. Should we leave them when printing or take them out?


src/ENDFtk/file.hpp
```diff
@@ -36,6 +36,50 @@ class Type {
   hasMT( int sectionNo ) const { return this->sectionMap.count( sectionNo ); }
   bool
   hasSectionNumber( int sectionNo ) const {  return this->hasMT( sectionNo ); }
+
+  int MF() const { return FileNumber; }
+
+#include "ENDFtk/file/Type/src/print.hpp"
+};
+
+/* MF1 has a limited number of possible sections */
```
apmccartney: Please move this to it's own file.

whaeck: I currently put it in #include "ENDFtk/file/file-1.hpp" while keeping the ctr-1.hpp and print-1.hpp in the original location "ENDFtk/file/Type/src". Is that OK, or should I put them in "ENDFtk/file/Type-1/src" and rename them ctr.hpp and print.hpp?

I can still move the files around - I have not pushed yet.

PR #39, opened 2017-10-16
Feature/mf13

jlconlin:


src/ENDFtk/section/13/Subsection.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include "ENDFtk/section/13/Subsection/src/ctor.hpp"
+
+  double EG() const { return this->table.C1(); }
+  double EK() const { return this->table.C2(); }
+  int LP() const { return this->table.L1(); }
+  int LF() const { return this->table.L2(); }
+  auto energies() const { return this->table.x(); }
+  auto crossSections() const { return this->table.y(); }
+
```
apmccartney: please encapsulate `table` and add a const reference getter `TAB1`

src/ENDFtk/section/13.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+  auto energies() const { return subsections_.front().x(); }
+  auto crossSections() const { return subsections_.front().y(); }
+  auto subsections() const {
+    return
+      this->subsections_
+      | ranges::view::drop_exactly( this->NK() > 1 )
+      | ranges::view::transform
+        ( []( const auto& tabulationRecord ) -> const Subsection&
+          { return static_cast< Subsection& >( tabulationRecord ); } );
```
apmccartney:
```cpp
{ return static_cast< Subsection& >( tabulationRecord ); } );
```

to

```cpp
{ return static_cast< const Subsection& >( tabulationRecord ); } );
```

src/ENDFtk/section/13.hpp
```diff
@@ -0,0 +1,30 @@
...
+  }
+};
```
apmccartney: still needs `NC` (which provides line count) and `print` methods

src/ENDFtk/section/13/Subsection.hpp
```diff
@@ -0,0 +1,13 @@
...
+public:
+  #include "ENDFtk/section/13/Subsection/src/ctor.hpp"
+
+  double EG() const { return TabulationRecord::C1(); }
+  double EK() const { return TabulationRecord::C2(); }
+  int LP() const { return TabulationRecord::L1(); }
+  int LF() const { return TabulationRecord::L2(); }
+  auto energies() const { return TabulationRecord::x(); }
+  auto crossSections() const { return TabulationRecord::y(); }
+};
```
apmccartney: `using TabulationRecord::print` and `using TabulationRecord::NC;`

src/ENDFtk/resonanceParameters/unresolved/EnergyDependent/test/EnergyDependent.test.cpp
```diff
@@ -28,63 +28,141 @@ SCENARIO( "Testing energy-dependent unresolved resonances" ){
           base, begin, end, lineNumber, MAT, MF, MT );
 
       THEN( "the parameters can be verified" ){
-        REQUIRE( Approx(6E3) == ted.EL() );
-        REQUIRE( Approx(1E5) == ted.EH() );
-        REQUIRE( 0 == ted.NRO() );
-        REQUIRE( 0 == ted.NAPS() );
-        REQUIRE( 80 == ted.NC() );
+        CHECK( Approx(6E3) == ted.EL() );
```
whaeck: Why change this from REQUIRE to CHECK? We've always done it this way.

src/ENDFtk/section/13.hpp
whaeck: This was added.

src/ENDFtk/section/13/Subsection.hpp
whaeck: Added as well.

src/ENDFtk/section/13/Subsection/test/Subsection.test.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "catch.hpp"
+#include "ENDFtk.hpp"
+
+using namespace njoy::ENDFtk;
+std::string sSubSection();
+
+SCENARIO( "section::Type< 13 >::SubSection" ){
+  GIVEN( "a string representation of a valid MF=13 Subsection" ){
+    std::string sSS = sSubSection();
+    // njoy::Log::info("sSS:\n{}", sSS );
```
whaeck: I think you can remove this line

src/ENDFtk/resonanceParameters/unresolved/EnergyDependent/test/EnergyDependent.test.cpp
jlconlin: I know we've always done it with `REQUIRE`, but I think we should use `CHECK`. `CHECK` will continue execution if it fails while `REQUIRE` will not perform the other tests when it fails. We could be hiding later issues that will take longer to resolve when we use `REQUIRE` instead of `CHECK`.

src/ENDFtk/section/13/Subsection/test/Subsection.test.cpp
jlconlin: Removed

src/ENDFtk/resonanceParameters/unresolved/EnergyDependent/test/EnergyDependent.test.cpp
whaeck: Maybe, I am not convinced as the developer has to run the tests anyway.

whaeck:
We'll look at the coverage when we pull this into the mf6 branch, which is up next merge merge in master.

PR #51, opened 2019-04-16
Added file::Type<> constructor using a vector of section::Type<>

whaeck:


src/ENDFtk/file/Type/src/ctor.hpp
```diff
@@ -1,3 +1,12 @@
+Type( std::vector< Section > sections ) :
+  sectionVector( sections ),
+  sectionMap( collectMap( this->sectionVector ) ) {
+
+  ranges::sort( this->sectionVector,
```
apmccartney: Using a projection could save you from instantiating a new type

```cpp
ranges::sort(this->sectionVector, std::less<>{}, &Section::MT);
```

PR #57, opened 2019-05-31
Changing the structure of the syntaxTree.

jlconlin:
The `syntaxTree::Tape` now owns it's own buffer.


jlconlin:
This should be a very quick PR.

src/ENDFtk.hpp
```diff
@@ -4,6 +4,7 @@
 #include <functional>
 #include <unordered_map>
 #include <complex>
+#include <map>
```
apmccartney: TIL `std::multimap` is defined in the `map` header

apmccartney: Are we safe to remove the tsl hopscotch map now?

src/ENDFtk/syntaxTree/File/src/ctor.hpp
```diff
@@ -2,20 +2,26 @@ File( const HEAD& head, BufferIterator begin,
...
+  sections_( createMap( head, begin, position, end, lineNumber ) ),
   bufferLimits( { begin, position } ){}
 catch( std::exception& e ){
   Log::info( "Trouble encountered while constructing a file syntax tree." );
   Log::info( "File number (MF): {}", head.MF() );
   throw e;
 }
 
-File( const File& other ) :
+/*
```
apmccartney: I think we can safely delete this at this point.

src/ENDFtk/syntaxTree/Material/src/ctor.hpp
```diff
@@ -2,19 +2,25 @@ Material( const HEAD& head, BufferIterator begin, BufferIterator& position,
...
+  files_( createMap( head, begin, position, end, lineNumber ) ),
   bufferLimits( { begin, position } ){}
 catch( std::exception& e ){
   Log::info( "Trouble encountered while constructing a material syntax tree." );
   Log::info( "Material number (MAT): {}", head.MAT() );
   throw e;
 }
 
-Material( const Material& other ) :
+/*
```
apmccartney: I think this can be deleted

src/ENDFtk.hpp
jlconlin: hopscotch map has been removed.

jlconlin:
The `syntaxTree` parser bug has been resolved as well as the other Pull Request comments. It should be ready to go.

src/ENDFtk/syntaxTree/File/src/createMap.hpp
```diff
@@ -1,17 +1,45 @@
...
 
-  for ( auto& section : vector ){
-    if ( map.count( section.MT() ) ){
+  std::map< int, Section_t > sections;
+
+  sections.emplace( 
+      head.MT(),
+      Section_t( head, begin, position, end, lineNumber ) );
+
+  begin = position;
```
whaeck: Lines 8-13: would it be possible to just say:
```
auto division = StructureDivision( head );
```
so that we can do the while loop immediately instead of having to do the first MF we find outside of it? Lines 8-12 and reading a new StructureDivision are essentially what is done in the while loop.

src/ENDFtk/syntaxTree/Material.hpp
```diff
@@ -30,23 +28,18 @@ class Material {
   MF( int fileNo ) { return this->fileNumber( fileNo ); }
 
   bool
-  hasMF( int fileNo ) const { return this->fileMap.count( fileNo ); }
+  hasMF( int fileNo ) const { return this->files_.count( fileNo ); }
 
   bool
   hasFileNumber( int fileNo ) const { return this->hasMF( fileNo ); }
```
whaeck: Some remark as above: file(...) and hasFile(...)?

src/ENDFtk/syntaxTree/Material/src/createMap.hpp
```diff
@@ -1,17 +1,41 @@
...
+createMap
+( const HEAD& head, BufferIterator begin,
+  BufferIterator& position, const BufferIterator& end, long& lineNumber ){
+
+  std::map< int, File_t > files;
+
+  files.emplace( head.MF(), File_t( head, begin, position, end, lineNumber ) );
+
+  begin = position;
+  auto division = StructureDivision( position, end, lineNumber );
```
whaeck: Same remark as above: line 8-11 are what is done in the while loop.

src/ENDFtk/syntaxTree/File/src/createMap.hpp
jlconlin: I like the idea, however:

```
error: cannot cast 'const HEAD' 
    (aka 'const njoy::ENDFtk::HeadRecord') to its protected base class
    'const njoy::ENDFtk::StructureDivision'
```
So it would take something more to make this work and I'd rather just get this completed.

src/ENDFtk/syntaxTree/File.hpp
```diff
@@ -31,18 +29,20 @@ class File {
   MT( int sectionNo ){ return this->sectionNumber( sectionNo ); }
```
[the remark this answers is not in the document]
jlconlin: While `section::Type<>` uses `sections(...)`,  `section::Base<>` uses `MT(...)`. Which one do you prefer?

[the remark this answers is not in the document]
apmccartney: Is this resolved?

PR #58, opened 2019-06-24
Feature/mf4

whaeck:
Full MF4 capability, no changes other than MF4 ;-)

Coverage should be of the order of 99% (I will add a coverage report once I get back to LANL).


src/ENDFtk/section/4.hpp
```diff
@@ -0,0 +1,90 @@
...
+  /* get methods */
+
+  /**
+   *  @brief Return the reference frame
+   */
+  int LCT() const { return this->lct_; }
+
+  /**
+   *  @brief Return the reference frame
+   */
```
apmccartney: An enum for this would be welcome

src/ENDFtk/section/4.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   *  @brief Return the distribution law
+   */
+  int LTT() const { return this->distributions_.index(); }
+
+  /**
+   *  @brief Return the distribution law
+   */
+  int LAW() const { return this->LTT(); }
```
apmccartney: Is LAW also an ENDF-specific name or was this intended as a 'hum-readable' alias to LTT? if the latter, need not scream

src/ENDFtk/section/4/MixedDistributions.hpp
```diff
@@ -0,0 +1,55 @@
...
+   *  @brief Return the Legendre distributions
+   */
+  const auto& legendre() const { return this->legendre_; }
+
+  /** 
+   *  @brief Return the tabulated distributions
+   */
+  const auto& tabulated() const { return this->tabulated_; }
+
+  /**
```
apmccartney: perhaps a bit of sugar?

```cpp
auto boundaries() const {
  return ranges::view::concat( this->legendre().boundaries(),
                              this->tabulated().boundaries() );
}

auto interpolants() const {
  return ranges::view::concat( this->legendre().interpolants(),
                                this->tabulated().interpolants() );
}

auto angularDistributions() const {
  using SumType = std::variant< std::reference_wrapper< const LegendreCoefficients >,
                                std::reference_wrapper< const TabulatedDistribution > >
  auto lift = ranges::view::transform( []( const auto& element ) -> SumType {
    return std::cref( element );
  } );
  return ranges::view::join( this->legendre().angularDistributions() | lift,
                             this->tabulated().angularDistributions() | lift );
}
```

src/ENDFtk/section/4/MixedDistributions/src/verifyBoundary.hpp
```diff
@@ -0,0 +1,13 @@
+static void
+verifyBoundary( double left, double right ) {
+
+  if ( left != right ) {
```
apmccartney: Note to self, double check this

apmccartney:
Double checked on the doublely-defined energy grid point. You were correct!

src/ENDFtk/section/4.hpp
whaeck: We'll have to do that in several places. Maybe do this globally for all cases like this in a separate pull request?

Also: these are essentially non-critical values that do not impact reading the file. Using an enum would allow us to check this when the user asks for the value (and not when we parse it).

src/ENDFtk/section/4.hpp
whaeck: I started doing this because Jeremy wanted all formatting options for angular and energy distributions called LAW as well. The term LAW is actually only defined in MF6. To be consistent with MF6, I have to scream :-)

src/ENDFtk/section/4/AngularDistributions.hpp
```diff
@@ -0,0 +1,62 @@
+/**
+ *  @class
+ *  @brief Base class for Legendre or tabulated angular distributions
+ */
+template < typename Records >
```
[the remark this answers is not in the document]
whaeck: Good catch, yes I think I can do that.

src/ENDFtk/section/4/MixedDistributions.hpp
whaeck: I like it, consider it added ;-)

whaeck:
I'll make the suggested changes. Once I commit those, we should be good to go.

src/ENDFtk/section/4.hpp
jlconlin: For now, let's leave this as it is and we can come back to this if needed.

src/ENDFtk/section/4/MixedDistributions/src/verifyBoundary.hpp
jlconlin: Double check what?

whaeck: Resolved in the next comment below.

PR #60, opened 2019-09-24
Feature/mf12

whaeck:
Full MF12 capability

While independent from the MF4 branch, the MF12 should be merged after the MF4 branch.


whaeck:
@jlconlin We're ready for this one ...

src/ENDFtk/section/12/PartialMultiplicity/src/ctor.hpp
```diff
@@ -0,0 +1,53 @@
...
+ */
+PartialMultiplicity( double energy,
+                     double level,
+                     int lp,
+                     int law,
+                     std::vector< long >&& boundaries,
+                     std::vector< long >&& interpolants,
+                     std::vector< double >&& cosines,
+                     std::vector< double >&& probabilities )
+  try : TabulationRecord( energy, level, lp, law,
```
jlconlin: How come there are no checks on the validity of `energy`, `level`, `lp`, and `law`?

src/ENDFtk/section/12/PartialMultiplicity/test/PartialMultiplicity.test.cpp
```diff
@@ -0,0 +1,169 @@
...
+                                             std::move( boundaries ),
+                                             std::move( wrongInterpolants ),
+                                             std::move( x ),
+                                             std::move( y ) ) );
+      } // THEN
+    } // WHEN
+
+    WHEN( "a string representation with an error is given" ) {
+
+      // no need to test every possibility (TAB1 takes care of tests)
```
jlconlin: 👍

src/ENDFtk/section/12/PartialMultiplicity/test/PartialMultiplicity.test.cpp
```diff
@@ -0,0 +1,169 @@
...
+  GIVEN( "invalid data for a PartialMultiplicity" ) {
+
+    WHEN( "inconsistent data is used" ) {
+
+      // no need to test every possibility (TAB1 takes care of tests)
+
+      double eg = 0.0;
+      double es = 3.0;
+      int lp = 2;
+      int lf = 1;
```
jlconlin: What about wrong values for these?

src/ENDFtk/section/12/TotalMultiplicity/src/ctor.hpp
```diff
@@ -0,0 +1,46 @@
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] boundaries       the interpolation range boundaries
+ *  @param[in] interpolants     the interpolation types for each range
+ *  @param[in] energies         the cosine values
+ *  @param[in] multiplicities   the probability values
+ */
+TotalMultiplicity( std::vector< long >&& boundaries,
```
jlconlin: I'd like to eventually see these templated as a generic `Range` type. Right now we are forcing people to use `std::vector<...>`.

src/ENDFtk/section/12/Multiplicities/test/Multiplicities.test.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+SCENARIO( "Multiplicities" ) {
+
+  GIVEN( "valid data for a Multiplicities with one partial" ) {
+
+    std::string string = chunkWithOnePartial();
+
+    WHEN( "the data is given explicitly" ) {
+
+      PartialMultiplicity partial = { 0.0, 0.0, 0, 1,
```
jlconlin: I'm not sure what the first four numbers in the constructor are for. Please explain

src/ENDFtk/section/12/TotalMultiplicity/test/TotalMultiplicity.test.cpp
```diff
@@ -0,0 +1,149 @@
...
+        CHECK( buffer == string );
+      } // THEN
+    } // WHEN
+  } // GIVEN
+
+  GIVEN( "invalid data for a TotalMultiplicity" ) {
+
+    WHEN( "inconsistent data is used" ) {
+
+      // no need to test every possibility (TAB1 takes care of tests)
```
jlconlin: 👍

src/ENDFtk/section/12/TransitionProbabilities.hpp
```diff
@@ -0,0 +1,126 @@
...
+  auto transitionProbabilities() const { return this->TP(); }
+
+  /**
+   *  @brief Return the conditional probabilities.
+   */
+  auto GP() const {
+
+    int lg = this->LG();
+    // need to do it this way to get the same return type
+    auto gp = [lg] ( const auto& chunk ) { return lg == 2 ? chunk[2] : 1.0; };
```
jlconlin: 👍

src/ENDFtk/section/12/TransitionProbabilities/src/ctor.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ *  @brief Constructor (for the simple case)
+ *
+ *  @param[in] energy          the highest energy level
+ *  @param[in] energies        the energy levels below the highest level
+ *                             (NT values)
+ *  @param[in] probabilities   the transition probabilities (NT values)
+ *  @param[in] conditionals    the conditional photon probabilities (NT values)
+ */
```
jlconlin: Like before, would like to see these templated on generic range type instead of `std::vector`.

src/ENDFtk/section/12/TransitionProbabilities/src/generateList.hpp
```diff
@@ -0,0 +1,48 @@
...
+static std::vector< double >
+generateList( std::vector< double >&& energies,
+              std::vector< double >&& probabilities,
+              std::vector< double >&& conditionalProbabilities ) {
+
+  unsigned int NT = energies.size();
+  if ( ( energies.size() != probabilities.size() ) or
+       ( energies.size() != conditionalProbabilities.size() ) ) {
+
+    Log::error( "The number of energy and probability values must be the same" );
```
jlconlin: Don't forget conditional probabilities~

src/ENDFtk/section/12/TransitionProbabilities/src/generateList.hpp
```diff
@@ -0,0 +1,48 @@
...
+  unsigned int NT = energies.size();
+  if ( energies.size() != probabilities.size() ) {
+
+    Log::error( "The number of energy and probability values must be the same" );
+    Log::info( "energies.size(): {}", NT );
+    Log::info( "probabilities.size(): {}", probabilities.size() );
+    throw std::exception();
+  }
+
+  std::vector< double > list;
```
jlconlin: You should declare this variable first thing in the function. For some reason that is required for return value optimization (RVO). You should be able to reserve the right amount of entries at this point.

src/ENDFtk/section/12/TransitionProbabilities/src/generateList.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+    Log::error( "The number of energy and probability values must be the same" );
+    Log::info( "energies.size(): {}", NT );
+    Log::info( "probabilities.size(): {}", probabilities.size() );
+    Log::info( "conditionalProbabilities.size(): {}",
+               conditionalProbabilities.size() );
+    throw std::exception();
+  }
+
+  std::vector< double > list;
```
jlconlin: Like above, move this to the first line in the function and reserve memory when you know how much you need.

src/ENDFtk/section/12/TransitionProbabilities/src/verifyLG.hpp
```diff
@@ -0,0 +1,11 @@
+static void
+verifyLG( int LG ) {
+
+  if ( ( LG != 1 ) and ( LG != 2 ) ) {
```
jlconlin: :+1

src/ENDFtk/section/12/TransitionProbabilities/test/TransitionProbabilities.test.cpp
```diff
@@ -0,0 +1,266 @@
...
+        chunk.print( output, 9228, 12, 51 );
+
+        CHECK( buffer == string );
+      } // THEN
+    } // WHEN
+  } // GIVEN
+
+  GIVEN( "invalid data for a TransitionProbabilities" ) {
+
+    WHEN( "inconsistent data is used for LG = 1" ) {
```
jlconlin: I'm no sure what is incosistent here.

src/ENDFtk/section/12/TransitionProbabilities/test/TransitionProbabilities.test.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+      THEN( "an exception is thrown" ) {
+
+        REQUIRE_THROWS( TransitionProbabilities( energy,
+                                                 std::move( energies ),
+                                                 std::move( probabilities ) ) );
+      } // THEN
+    } // WHEN
+
+    WHEN( "inconsistent data is used for LG = 2" ) {
```
jlconlin: I'm not sure what is inconsistent here

src/ENDFtk/section/12/TransitionProbabilities/test/TransitionProbabilities.test.cpp
```diff
@@ -0,0 +1,266 @@
...
+      auto end = string.end();
+      long lineNumber = 1;
+
+      THEN( "an exception is thrown" ) {
+
+        REQUIRE_THROWS( TransitionProbabilities( begin, end, lineNumber,
+                                                 9228, 12, 51 ) );
+      } // THEN
+    } // WHEN
+  } // GIVEN
```
jlconlin: Please check for invalid `LG` values.

src/ENDFtk/section/12/src/readPhotonProduction.hpp
```diff
@@ -0,0 +1,23 @@
...
+                      int MF,
+                      int MT,
+                      int LO,
+                      int N ) {
+
+  switch ( LO ) {
+    case 1: return Multiplicities( begin, end, lineNumber, MAT, MF, MT, N );
+    case 2: return TransitionProbabilities( begin, end, lineNumber, MAT, MF, MT );
+    default : {
+      Log::error( "Encountered illegal LO value" );
```
jlconlin: :+1

src/ENDFtk/section/12/src/readPhotonProduction.hpp
```diff
@@ -0,0 +1,23 @@
...
+                      const Iterator& end,
+                      long& lineNumber,
+                      int MAT,
+                      int MF,
+                      int MT,
+                      int LO,
+                      int N ) {
+
+  switch ( LO ) {
+    case 1: return Multiplicities( begin, end, lineNumber, MAT, MF, MT, N );
```
jlconlin: Does gcc give an error about not having a `break` statement?

src/ENDFtk/section/12/src/print.hpp
```diff
@@ -0,0 +1,40 @@
...
+ *  @param[in] MF            the MF number
+ */
+template< typename OutputIterator >
+void print( OutputIterator& it, int MAT, int MF ) const {
+
+  struct L2 {
+
+    int operator()( const Multiplicities& ) { return 0; }
+    int operator()( const TransitionProbabilities& value ) {
+
```
jlconlin: Removing this blank line makes it easier (in my opinion) to know that the next line belongs to this function. At first glance, it looks like it is something different.

jlconlin: That's just my opinion.

src/ENDFtk/section/12/src/print.hpp
```diff
@@ -0,0 +1,40 @@
...
+      return value.LG();
+    }
+  };
+
+  struct N1 {
+
+    int operator()( const Multiplicities& value ) { return value.NK(); }
+    int operator()( const TransitionProbabilities& value ) {
+
```
jlconlin: same comment about blank line

src/ENDFtk/section/12/test/12.test.cpp
```diff
@@ -0,0 +1,559 @@
...
+
+// convenience typedefs
+using Multiplicities =
+section::Type< 12 >::Multiplicities;
+using PartialMultiplicity =
+section::Type< 12 >::PartialMultiplicity;
+using TotalMultiplicity =
+section::Type< 12 >::TotalMultiplicity;
+using TransitionProbabilities =
+section::Type< 12 >::TransitionProbabilities;
```
jlconlin: I think some (all?) of these could be on one line.

src/ENDFtk/section.hpp
```diff
@@ -13,5 +13,6 @@ class Type;
 #include "ENDFtk/section/6.hpp"
 #include "ENDFtk/section/7.hpp"
 #include "ENDFtk/section/8.hpp"
+#include "ENDFtk/section/12.hpp"
```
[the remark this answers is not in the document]
whaeck: Why do you think this is wrong, or do you just have a bad sense of humour?

src/ENDFtk/section/12/Multiplicities.hpp
```diff
@@ -0,0 +1,68 @@
...
+ *  as multiplicities. The data consists of a range of partial multiplicities
+ *  and an optional total multiplicity (if there is more than 1 partial).
+ *
+ *  See ENDF102, section 12.2.1 for more information.
+ */
+class Multiplicities {
+
+  /* fields */
+  std::optional< TotalMultiplicity > total_;
+  std::vector< PartialMultiplicity > partials_;
```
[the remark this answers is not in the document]
whaeck: They are essentially different because the total multiplicity does not define the variables EG, ES, LP and LF (which only make sense for the partial multiplicity). In some cases, a value of zero is actually undefined so I did not want to reuse the partial multiplicity interface for the total.

src/ENDFtk/section/12/Multiplicities/src/ctor.hpp
```diff
@@ -0,0 +1,74 @@
...
+ *  @param[in] MAT          the expected MAT number
+ *  @param[in] MF           the expected MF number
+ *  @param[in] MT           the expected MT number
+ *  @param[in] NK           the number of partials
+ */
+template< typename Iterator >
+Multiplicities( Iterator& begin, const Iterator& end,
+                long& lineNumber, int MAT, int MF, int MT,
+                int NK )
+  try : Multiplicities( readTotalMultiplicity( begin, end, lineNumber,
```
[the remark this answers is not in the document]
whaeck: It is the first, it says so in the manual in section 12.2.1:
```
[MAT, 12, MT/ ZA, AWR, LO, 0, NK, 0]HEAD (LO=1)
[MAT, 12, MT/ 0.0, 0.0, 0, 0, NR, NP/Eint / Y (E)] TAB1 (if NK>1)
<subsection for k =1>
<subsection for k=2>
......
<subsection for k=NK>
[MAT, 12, 0/ 0.0, 0.0, 0, 0, 0, 0] SEND
```
The first TAB1 here is the total multiplicity (only given if NK>1). The subsections that follow are the partial multiplicities:
```
[MAT, 12, MT/ Egk, ESk, LP, LF, NR, NP/ Eint / yk(E)] TAB1
```

src/ENDFtk/section/12/Multiplicities/test/Multiplicities.test.cpp
```diff
@@ -0,0 +1,282 @@
...
+void verifyChunkWithOnePartial( const Multiplicities& );
+std::string chunkWithMultiplePartials();
+void verifyChunkWithMultiplePartials( const Multiplicities& );
+std::string invalidChunk();
+
+SCENARIO( "Multiplicities" ) {
+
+  GIVEN( "valid data for a Multiplicities with one partial" ) {
+
+    std::string string = chunkWithOnePartial();
```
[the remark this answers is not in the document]
whaeck: I also use it inside ```WHEN( "the data is read from a string/stream" )```, which is why I define it outside the WHENs in the test. I have consistently used this construct in all tests in ENDFtk.

src/ENDFtk/section/12/PartialMultiplicity/src/ctor.hpp
whaeck: So none of these values are essential in the format: if the values are "wrong", they will not impact the reading of the file. It is questionable to end reading the file for the simple reason that an enumerated value is wrong. This may be better handled in a function that looks for these things specifically once we have read the file (and if the user asks for it).

It's a valid point however, especially when we create these components from scratch. A solution would be to accept not an integer but a struct/enum that would check the value before we give it to this constructor. The interface returning the value would then return the same struct, providing a runtime exception when you request the value instead of when you read the file.

src/ENDFtk/section/12/PartialMultiplicity/test/PartialMultiplicity.test.cpp
whaeck: See above.

src/ENDFtk/section/12/TotalMultiplicity/src/ctor.hpp
whaeck: I can go back and do that everywhere else as well, but that'll add at least a month of work to change the interfaces, change the implementations and the tests. You really want to go there?

whaeck: However: I agree that it may be useful, but not now.

src/ENDFtk/section/12/Multiplicities/test/Multiplicities.test.cpp
whaeck: level energy, photon or binding energy (depends on LP), LP and LF

src/ENDFtk/section/12/TransitionProbabilities/src/ctor.hpp
whaeck: see above

src/ENDFtk/section/12/TransitionProbabilities/src/generateList.hpp
whaeck: I don't grasp thy meaning ...

src/ENDFtk/section/12/TransitionProbabilities/src/generateList.hpp
whaeck: I have heard that, I tried finding if it is actually a requirement. I know Austin said so at some point. In this case, I test before actually creating an object. It's a small change anyway.

src/ENDFtk/section/12/TransitionProbabilities/test/TransitionProbabilities.test.cpp
whaeck: inconsistent size for the various arrays, I added a comment (I normally indicate this in the tests but I appear not to have done that here)

src/ENDFtk/section/12/TransitionProbabilities/test/TransitionProbabilities.test.cpp
whaeck: inconsistent size for the various arrays, I added a comment (I normally indicate this in the tests but I appear not to have done that here)

src/ENDFtk/section/12/TransitionProbabilities/test/TransitionProbabilities.test.cpp
whaeck: So, when constructing from scratch, the values of LG have not to be given as they are determined from the case we're looking at. The value of LG does not impact reading the lists either, as it only impacts the size. In addition, 

I could add a verifyLG to see if it is 1 or 2, as in this case it influences interpretation of the list?

src/ENDFtk/section/12/src/readPhotonProduction.hpp
whaeck: Putting a break after a return is unreachable.

src/ENDFtk/section/12/src/print.hpp
whaeck: I have been doing this consistently: new line after opening brace.

src/ENDFtk/section.hpp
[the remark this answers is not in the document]
jlconlin: No YOU have a bad sense of humor

src/ENDFtk/section/12/TotalMultiplicity/src/ctor.hpp
jlconlin: Agreed. It would be nice, but let's not go change it right now.

When we get around to doing the next section, let's use a generic `Range`.

src/ENDFtk/section/12/TransitionProbabilities/src/generateList.hpp
jlconlin: It would be nice if the error message explicitly mentioned both probabilities and conditional probabilities. Right now it only mentions the size of the probabilities. Perhaps it's too much information to give to the user?

whaeck: Conditional probabilities are still probability values ...

src/ENDFtk/section/12/TotalMultiplicity/src/ctor.hpp
whaeck: I would actually do this in one fell swoop on the entire library instead of gradually adding it section by section.

src/ENDFtk/section/12/TransitionProbabilities/src/generateList.hpp
whaeck: The information lines actually provide the sizes of all of them.

src/ENDFtk/section/12/TransitionProbabilities/test/TransitionProbabilities.test.cpp
whaeck: My bad: it actually already exists, so yes, I need to test for LG values.

whaeck: I'm getting confused: so TransitionProbabilities does test LG values through the size of the array by dividing the size with NT. In the invalidChunk(), NT=1 with the size equal to 6, as such LG=6 which is invalid. I added a comment to highlight this.

whaeck: I also added a test with a list using NPL=4 while the number of values in the list is actually 6 (should be picked up by the LIST constructor)

src/ENDFtk/section/12/TransitionProbabilities/Transition.hpp
```diff
@@ -0,0 +1,51 @@
...
+ *  This chunk contains the energy of the transition, the transitional
+ *  probability and the conditional probability (if it is defined).
+ *
+ *  See ENDF102, section 12.2.2 for more information.
+ */
+template < typename Chunk >
+class Transition {
+
+  /* fields */
+  Chunk chunk;
```
jlconlin: So I like separating this out, but `Chunk` is a bit opaque and *very* arbitrary. I could pass in a vector of size==1000 and this would compile just fine. We could use some SFINAE to restrict what could be used for `Chunk`, but I'm not sure I care that much. However, a more descriptive name would be helpful.

whaeck: Well, what I'll be getting is a result from ranges::view::chunk(), hence the name for the template argument. The class is actually called Transition, which is what the Chunk I'm getting is going to be interpreted as.

About the SFINAE piece of things: should we really provide such a feature on an object that we only create internally, and that is not used in any of the public constructors (and thus relatively useless to a user)?

The internals of the TransitionProbabilities class will ensure that the Chunk is a size 2 or 3 chunk. It is impossible for this to be of any other size by default due to the state checking we've implemented.

PR #61, opened 2019-11-05
Feature/mf13

whaeck:
Photon production cross sections in MF13: refactored and expanded with constructors to make it from scratch. Even though the ENDF manual says "the structure is very similar to MF12", it is not (e.g. there is no variant in this one).

MF12 nomenclature and interface for similar components are the same here.


src/ENDFtk/section/13/PartialCrossSection/src/ctor.hpp
```diff
@@ -0,0 +1,55 @@
...
+ *  @param[in] level            the energy of the level from which the photon
+ *                              originates
+ *  @param[in] lp               the primary photon flag (0, 1 or 2)
+ *  @param[in] law              the photon energy distribution law
+ *  @param[in] boundaries       the interpolation range boundaries
+ *  @param[in] interpolants     the interpolation types for each range
+ *  @param[in] energies         the energy values
+ *  @param[in] xs               the partial cross section values
+ */
+PartialCrossSection( double energy,
```
jlconlin: Isn't this the same constructor as for the `TabulationRecord`? Do we need to duplicate it here? If the answer is "so we can provide a Log message", that's fine. I'm just trying to understand what we are doing.

src/ENDFtk/section/13/TotalCrossSection.hpp
```diff
@@ -0,0 +1,44 @@
+/**
+ *  @class
+ *  @brief The total photon production cross section.
+ *
+ *  See ENDF102, section 13.2 for more information.
+ */
+class TotalCrossSection : protected TabulationRecord {
```
jlconlin: Why do we need a separate class for this? According to the ENDF documentation, all the MF=13 subsections have the same format. 

I think we've had this discussion, but I can't remember what we talked about.

src/ENDFtk/section/13/src/NC.hpp
```diff
@@ -1,5 +1,10 @@
...
+ */
 long NC() const {
+
   long NC = 1;
-  for ( auto& entry : this->subsections_ ) { NC += entry.NC(); }
+  if ( this->total_ ) { NC += this->total_->NC(); }
+  for ( const auto& entry : this->partials_ ) { NC += entry.NC(); }
   return NC;
-} 
+};
```
jlconlin: This semi-colon is not needed. I believe gcc will get upset at this. (I'm checking.)

src/ENDFtk/section/13/src/ctor.hpp
```diff
@@ -1,28 +1,87 @@
...
+  total_( std::move( total ) ),
+  partials_( std::move( partials ) ) {
+
+  verifyNK( this->NK() );
+}
+
+public:
+
+/**
+ *  @brief Constructor (only one partial given, no total required)
```
jlconlin: If there is only one partial given, doesn't that imply that it is the total?

src/ENDFtk/section/13/src/print.hpp
```diff
@@ -1,11 +1,22 @@
...
+ *
+ *  @tparam OutputIterator   an output iterator
+ *
+ *  @param[in] it            the current position in the output
+ *  @param[in] MAT           the MAT number
+ *  @param[in] MF            the MF number
+ */
 template< typename OutputIterator >
-void print( OutputIterator& it, int MAT ) const {
+void print( OutputIterator& it, int MAT, int MF ) const {
```
jlconlin: Why do we take `MF` as an input here? We know that it is 13 so why should we change it?

src/ENDFtk/section/13/PartialCrossSection/src/ctor.hpp
whaeck: That, and providing proper interface to extract the information. These are energy and cross section values not x and y values. It allows us to use the proper ENDF speak in every context.

src/ENDFtk/section/13/TotalCrossSection.hpp
whaeck: No, a total cross section is not the same as a partial cross section. It's C1, C2, L1 and L2 values are not defined as data and are 0 by default. That is not the case for the partial cross section. Again, this is about interface.

src/ENDFtk/section/13/src/NC.hpp
whaeck: If gcc got upset, I would have seen it ;-)

src/ENDFtk/section/13/src/print.hpp
whaeck: Because that's how it is done everywhere else. If we change it here, we should change it everywhere.

src/ENDFtk/section/13/src/ctor.hpp
whaeck: Yes. The only reason this signature is being used is to distinguish between the situation where you have only one partial (and thus do not have to use a total cross section) and the case where there is more than one partial so that the total must be given.
In the first case, the section to be used is still a partial cross section due to the additional information on the partial cross section still being required.

PR #63, opened 2020-02-11
Changing syntax tree template parameters.

jlconlin:
This is intended to make ENDFtk a little more consistent across the different syntaxTree classes (`Tape`, `Material`, `File`, and `Section`).


jlconlin:
> I'm getting ready to put in a PR for some other syntaxTree shenanigans, so let's get this reviewed. I'll start now.

I'm glad that you will take a look. I've been begging @whaeck to take a look for months now!

whaeck:
We have talked about this before but it wasn't urgent. And I still think having Buffer template argument on the class but use BufferIterator arguments in the constructor is confusing.

jlconlin:
I've certainly needed it these last couple of months.

whaeck:
No, you do not "need" it nor have you "needed" it. You could have continued to use an iterator type as template argument and it would just have worked as well.

Tape uses the buffer type because that's what a user should be using in the first place. Materials use iterators from that Buffer since they are used under the hood in Tape. A user should not be creating Material objects in the first place. The only reason we have to do so now, is that things are incomplete in ENDFtk.

nathangibson14:
I didn't notice the conversation between you and Wim while I was writing my review comment.  I guess I'm probably saying the same thing as Wim, but I am missing some of the history of your discussions.  What is the application you needed this change in recently?  I could certainly be missing something here.

jlconlin:
The main purpose for this is to make it easier to declare a type of `syntaxTree` `Tape`, `Material`, `File`, and `Section`. It would be convenient if they were similar, i.e.:
- `Tape< std::string >`
- `Material< std::string>`
- `File< std::string >`
- `Section< std::string >`
instead of the way they are prior to this Pull Request:
- `Tape< std::string >`
- `Material< std::string::iterator >`
- `File< std::string::iterator >`
- `Section< std::string::iterator >`

I am storing some `syntaxTree::Material` and needed a clean way to declare the variable type. This Pull Request makes things simpler and more consistent.

I am using a `syntaxTree::Material` in  [RECONR](https://github.com/njoy/RECONR) and was struggling to know how to declare the type. I had seen some examples of creating a `syntaxTree::Tape` and tried to do the same for `syntaxTree::Material`, but to no avail. As I looked further, I recognized the inconsistency and thought it would be pretty straightforward to make them all consistent.

I apologize to @whaeck, I did not mean to accuse him of neglect. I just wanted to state that I've wanted this for awhile.

whaeck:
If we wish to use specific syntax tree components by itself more safely, it may be wiser to make a more fundamental change to the syntax tree components by cutting up the buffer and giving it to the smallest component instead.

It would allow the following:
```
std::string content = ...
mat = syntaxTree::makeMaterial( std::move( content ) );
```
or:
```
std::string content = ...
section = syntaxTree::makeSection( std::move( content ) );
```

The question is if this is actually useful since 99% of ENDFtk users will have a tape to work with.

nathangibson14:
I'm still a little confused as to the impetus behind this change.  I assume RECONR is using the `syntaxTree` to read in a `Tape`, then extract and parse data from it?  Then, it should only need to instantiate `Tape< std::string >` and be able to use `auto mytape.materials()` to get a `Material`.  I found an instance of a `typedef` in what I could see of the RECONR implementation of `using Material_t = Material< std::string::iterator >` -- is it just that sort of call that you are wanting to shorten?  I suppose you might be using a `Material` as an input argument for function, rather than needing multiple parsed sections?

Again, I don't really have a dog in this fight, I don't see a big benefit either way.  To me, it makes sense that `Tape` is different from other `syntaxTree` components, as it's the only one instantiated with a buffer, and I imagine also the only one instantiated by a user.  But I can also see a little bit of missing intuition with the template arguments not all being the same.

To me, this seems like a bigger issue of lacking documentation.  I imagine you tried to use intuition and got an error, and are thus improving the intuitive aspect.  But all could be fixed with some documentation so that intuition isn't required! :)

I'll click the approve button once I run the test suite (got distracted this morning...), then I'll abstain from voting on merging it in.

whaeck:
I'm getting a compilation error with clang 11 (default Catalina xcode):
In file included from /Users/wim/Projects/njoy21/ENDFtk/src/ENDFtk/file/2/test/2.test.cpp:4:
In file included from /Users/wim/Projects/njoy21/ENDFtk/src/ENDFtk.hpp:19:
In file included from /Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/all.hpp:17:
In file included from /Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/core.hpp:18:
In file included from /Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/distance.hpp:22:
In file included from /Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/range_traits.hpp:24:
/Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/range_concepts.hpp:78:45: error: no matching function for call to object of type 'const _begin_::fn'
                using iterator_t = decltype(begin(std::declval<T &>()));
                                            ^~~~~
/Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/range_traits.hpp:35:9: note: in instantiation of template type alias 'iterator_t' requested here
        using iterator_t = concepts::Range::iterator_t<Rng>;
        ^
/Users/wim/Projects/njoy21/ENDFtk/src/ENDFtk/syntaxTree/File.hpp:5:34: note: in instantiation of template type alias 'iterator_t' requested here
  using BufferIterator = ranges::iterator_t< Buffer >;
                                 ^
/Users/wim/Projects/njoy21/ENDFtk/src/ENDFtk/file/2/test/2.test.cpp:141:9: note: in instantiation of template class 'njoy::ENDFtk::syntaxTree::File<std::__1::__wrap_iter<char *> >' requested here
        fileTree( asHead( division ), start, begin, end, lineNumber );
        ^
/Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/begin_end.hpp:88:32: note: candidate template ignored: substitution failure [with R = std::__1::__wrap_iter<char *>]: no matching function for
      call to 'impl_'
                constexpr auto operator()(R &r) const
                               ^
/Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/begin_end.hpp:96:32: note: candidate template ignored: substitution failure [with R = std::__1::__wrap_iter<char *>]: no matching function for
      call to 'impl_'
                constexpr auto operator()(const R &&r) const
                               ^
/Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/begin_end.hpp:103:32: note: candidate template ignored: could not match 'reference_wrapper' against '__wrap_iter'
                constexpr auto operator()(std::reference_wrapper<T> ref) const
                               ^
/Users/wim/Projects/njoy21/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/begin_end.hpp:110:32: note: candidate template ignored: could not match 'reference_wrapper' against '__wrap_iter'
                constexpr auto operator()(ranges::reference_wrapper<T> ref) const
                               ^
1 error generated.
make[2]: *** [src/ENDFtk/file/2/test/CMakeFiles/ENDFtk.file.2.test.dir/2.test.cpp.o] Error 1
make[1]: *** [src/ENDFtk/file/2/test/CMakeFiles/ENDFtk.file.2.test.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

jlconlin:
Yeah, I found two tests that I forgot to update. I just pushed two new changes a few minutes ago.

jlconlin:
@whaeck and I found that there is already an alias to get what I wanted: `syntaxTree::Tape< Buffer >::Material_t` (and similar ones for `File_t`, and `Section_t`. This meets my needs for now so this Pull Request is no longer needed.

PR #64, opened 2020-02-18
C++17

jlconlin:


metaconfigure/fetch_subprojects.py
```diff
@@ -30,35 +30,35 @@ def clone_submodule( relative_path ):
       ----------------------------------------
       """.format( relative_path = relative_path ) ) )
     invocation = [ "git", "submodule", "update", "-q","--init", "--", relative_path ]
-    if os.name == "nt":
-        invocation.insert( 0, "powershell" )
+#    if os.name == "nt":
+#        invocation.insert( 0, "powershell" )
```
whaeck: Why comment out code?

metaconfigure/generate.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python3
+#! /usr/bin/env python
```
whaeck: This may be a bad idea. python does not necessarily alias to python 3.4+

metaconfigure/fetch_subprojects.py
jlconlin: This is part of metaconfigure. We should address this in that repository.

metaconfigure/generate.py
jlconlin: This is also part of metaconfigure. We had a discussion about this awhile ago. I'm not sure we came to the right conclusion.

PR #65, opened 2020-03-17
Feature/mf2

whaeck:
Full MF2 MT151 functionality, made MF2 an enumerated file like MF7.

Some testing could be improved but overall coverage is good. Testing will be completed when adding MF2 MT152 and MF2 MT153 which are NJOY only sections added to the PENDF file by RECONR and PURR.


jlconlin:
Going through the Pull Request. However, I can't compile as a file is missing.

```console
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/section/2/151.hpp:16:12: fatal error: 'ENDFtk/section/2/151/src/verifyNIS.hpp' file not
      found
  #include "ENDFtk/section/2/151/src/verifyNIS.hpp"
```

src/ENDFtk/file/2.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+  /* fields */
+  Map sectionMap;
+
+public :
+
+  #include "ENDFtk/file/2/src/ctor.hpp"
+
+  bool
+  hasSection( int sectionNo ) const {
```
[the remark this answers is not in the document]
whaeck: Because sectionMap is not a normal map.
Yes, we'll have to make additional changes when we want to add 152 and 153. They are relatively easy to do, so I can do those now as well.

jlconlin:
Still can't compile everything:

```console
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceParameters/test/ResonanceParameters.test.cpp:81:32: warning:
      empty parentheses interpreted as a function declaration [-Wvexing-parse]
      ResonanceParameters chunk(  );
                               ^~~~
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceParameters/test/ResonanceParameters.test.cpp:81:32: note:
      replace parentheses with an initializer to declare a variable
      ResonanceParameters chunk(  );
                               ^~~~
                               {}
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceParameters/test/ResonanceParameters.test.cpp:81:27: error:
      functions that differ only in their return type cannot be overloaded
      ResonanceParameters chunk(  );
      ~~~~~~~~~~~~~~~~~~~ ^
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceParameters/test/ResonanceParameters.test.cpp:12:13: note:
      previous declaration is here
std::string chunk();
~~~~~~~~~~~ ^

```

src/ENDFtk/file/2/src/print.hpp
```diff
@@ -0,0 +1,10 @@
+template< typename OutputIterator >
+void print( OutputIterator& it, int MAT ) const {
+  this->sectionMap[ 151_c ].print( it, MAT, MF() );
+  /*
```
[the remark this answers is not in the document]
whaeck: We probably can do that, since optionalSections is currently empty.

src/ENDFtk/resonanceParameters/Isotope/test/Isotope.test.cpp
```diff
@@ -7,50 +7,58 @@ using namespace njoy::ENDFtk;
...
-      REQUIRE( 1.276553 == sc.AP() );
+      CHECK( 1 == iso.resonanceRanges().size() );
+
+      const auto& range = iso.resonanceRanges().back();
+      CHECK( 1E-5 == Approx( range.EL() ) );
+      CHECK( 1E-5 == Approx( range.lowerEnergy() ) );
+      CHECK( 1E+5 == Approx( range.EH() ) );
+      CHECK( 1E+5 == Approx( range.upperEnergy() ) );
+
+      CHECK( 0 == range.LRU() );
```
jlconlin: Nice job filling out all these tests.

src/ENDFtk/resonanceParameters/Isotope/test/Isotope.test.cpp
```diff
@@ -132,13 +188,13 @@ SCENARIO( "Isotope" ){
...
 
-      REQUIRE_THROWS(
+      CHECK_THROWS(
         resonanceParameters::Isotope( begin, end, lineNumber, MAT, MF, MT )
       );
     }
 
     WHEN( "there is one Rmatrix-Limited range in the Isotope" ){
-      std::string ENDF =
+/*      std::string ENDF =
```
jlconlin: Did you intend to leave this commented out?

src/ENDFtk/resonanceParameters/Isotope/test/Isotope.test.cpp
```diff
@@ -190,56 +246,56 @@ SCENARIO( "Isotope" ){
...
-        std::string ENDF = 
+        std::string ENDF =
           " 1.001000+3 1.000000+0          0          0          1          0 125 2151    2\n"
           " 1.000000-5 1.000000+5          1          5          0          0 125 2151    3\n";
         auto begin = ENDF.begin();
         auto end = ENDF.end();
         long lineNumber = 0;
 
-        REQUIRE_THROWS( 
+        CHECK_THROWS(
```
jlconlin: These `REQUIRE` to `CHECK` changes make me happy.

src/ENDFtk/resonanceParameters/ResonanceRange/src/ctor.hpp
```diff
@@ -0,0 +1,27 @@
...
+ *                                the scattering radius (0 or 1 if no energy
+ *                                dependent scattering radius is given and 0, 1
+ *                                or 2 if energy dependent scattering radius is
+ *                                given)
+ *  @param[in] parameters         the resonance parameters
+ *  @param[in] scatteringRadius   An optional energy dependent scattering radius
+ */
+ResonanceRange( double el, double eh, int naps,
+                ResonanceParameters&& parameters,
+                std::optional< ScatteringRadius > scatteringRadius = std::nullopt )
```
jlconlin: Should this be taken by r-value reference? You do `move` I on line 19.

src/ENDFtk/resonanceParameters/Isotope.hpp
```diff
@@ -1,21 +1,68 @@
...
+
+  /* constructor */
+  #include "ENDFtk/resonanceParameters/Isotope/src/ctor.hpp"
+
+  /* get methods */
+
+  /**
+   *  @brief Return the ZA identifier of the isotope
+   */
+  double ZAI() const { return this->zai_; }
```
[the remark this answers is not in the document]
whaeck: We don't have a human readable name for ZA on the head record either. As such, I did not provide one here for ZAI.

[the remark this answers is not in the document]
whaeck: I always try to provide a human readable name for ENDF variables.

src/ENDFtk/resonanceParameters/Isotope/test/Isotope.test.cpp
whaeck: Yes. For now.

src/ENDFtk/file/2/test/2.test.cpp
```diff
@@ -0,0 +1,243 @@
...
+void verifyChunk151( const file::Type< 2 >& );
+std::string validSEND();
+std::string validFEND();
+std::string validMEND();
+std::string validTEND();
+std::string validHEAD();
+
+SCENARIO( "Testing special case of file 2" ) {
+
+  GIVEN( "valid data for a File 1" ) {
```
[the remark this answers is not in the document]
whaeck: Yep.

src/ENDFtk/file/2/test/2.test.cpp
```diff
@@ -0,0 +1,243 @@
...
+    StructureDivision division( begin, end, lineNumber );
+    file::Type< 2 > file2( division, begin, end, lineNumber );
+
+    THEN( "it can be printed" ) {
+
+      std::string buffer;
+      auto output = std::back_inserter( buffer );
+      file2.print( output, 1122 );
+
+std::ofstream test( "orig.txt" );
```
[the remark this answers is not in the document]
whaeck: That's because that's not supposed to be there anymore.

src/ENDFtk/file/2/test/2.test.cpp
```diff
@@ -0,0 +1,243 @@
...
+      std::string twice( file2string.begin(), file2string.end() - 81 );
+      twice += file2string;
+      auto begin = twice.begin();
+      auto end = twice.end();
+      long lineNumber = 0;
+      StructureDivision division( begin, end, lineNumber );
+
+      THEN( "an exception is thrown" ) {
+
+        CHECK_THROWS( file::Type< 2 >
```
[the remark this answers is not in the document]
whaeck: We can't test on the exception type since all exceptions are std::exception. I did verify the logging messages.

src/ENDFtk/resonanceParameters/ResonanceRange/src/ctor.hpp
whaeck: Yes, a mere oversight on my part.

src/ENDFtk/file/2/test/2.test.cpp
```diff
@@ -0,0 +1,243 @@
...
+std::string validTEND();
+std::string validHEAD();
+
+SCENARIO( "Testing special case of file 2" ) {
+
+  GIVEN( "valid data for a File 1" ) {
+
+    WHEN( "a file::Type< 2 > is constructed using only mt151" ) {
+
+      // section::Type< 2, 151 > mt451( /* complete this */ );
```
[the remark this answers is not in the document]
whaeck: So do most other file tests. I'll do that later.

src/ENDFtk/resonanceParameters/Isotope.hpp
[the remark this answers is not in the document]
jlconlin: I know you try to do create a human readable name, that's why I was surprised that you didn't have one here.

src/ENDFtk/resonanceParameters/resolved/BreitWignerReichMooreBase/src/verifySize.hpp
```diff
@@ -0,0 +1,11 @@
+static void
+verifySize( int NLS ) {
+
+  if ( NLS == 0 ) {
```
jlconlin: Why not test `NLS < 1`?

src/ENDFtk/resonanceParameters/resolved/BreitWignerReichMooreLValueBase.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+  *  @brief Return the spin values
+   */
+  auto spinValues() const { return this->AJ(); }
+
+  /**
+   *  @brief Return the first width
+   */
+  auto G1() const {
```
jlconlin: Would it make sense to have a private function like:

```cpp
auto G( int drop ) const { 
return ListRecord::list() | ranges::view::drop_exactly( drop )
                          | ranges::view::stride( 6 );
}
```

Then al the `G#` functions could just call down to the private `G` function.

Or perhaps make a MACRO out of this?

src/ENDFtk/resonanceParameters/resolved/BreitWignerReichMooreLValueBase/src/generateList.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+  std::vector< double > list;
+  for ( unsigned int i = 0; i < nrs; ++i ) {
+
+    list.push_back( energies[i] );
+    list.push_back( spins[i] );
+    list.push_back( column1[i] );
+    list.push_back( column2[i] );
+    list.push_back( column3[i] );
+    list.push_back( column4[i] );
```
jlconlin: Should you `move` these values?

src/ENDFtk/resonanceParameters/resolved/BreitWignerLValue/Resonance.hpp
```diff
@@ -0,0 +1,90 @@
+/**
+ *  @class
+ *  @brief Convenience interface for Breit-Wigner resonances
+ *
+ *  See ENDF102, section 2.2.1.1 for more information.
+ */
+template < typename Range >
+class Resonance {
```
jlconlin: So, I couldn't see where this class was used. Can you point me to where it is?

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ParticlePairs.hpp
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   *  @brief Return the mass of the first particle in each particle pair
+   */
+  auto massParticleA() const { return this->MA(); }
+
+  /**
+   *  @brief Return the mass of the second particle in each particle pair
+   */
+  auto MB() const {
+    return ListRecord::list()
```
jlconlin: Like previous comment, would it make sense to write a private function to `drop_exactly` and `stride( 12 )` that each of these could then call? It's less direct, but also less code.

jlconlin: Also, what happens when you don't have enough numbers here? I assume `ranges` will throw some sort of exception. Do we need to `catch` it or do we ensure elsewhere that we have enough numbers?

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ParticlePairs.hpp
```diff
@@ -0,0 +1,184 @@
...
+             | ranges::view::stride( 12 ); }
+
+  /**
+   *  @brief Return the penetrability flag for each particle pair
+   */
+  auto PNT() const {
+    return ListRecord::list()
+             | ranges::view::drop_exactly( 7 )
+             | ranges::view::stride( 12 )
+             | ranges::view::transform( [] ( auto pnt )
```
jlconlin: If you did write another function, you could still transform on whatever it returns.

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ParticlePairs/src/generateList.hpp
```diff
@@ -0,0 +1,51 @@
+static std::vector< double >
+generateList( std::vector< double >&& ma, std::vector< double >&& mb,
+              std::vector< double >&& za, std::vector< double >&& zb,
+              std::vector< double >&& ia, std::vector< double >&& ib,
+              std::vector< double >&& q, std::vector< int >&& pnt,
+              std::vector< int >&& shf, std::vector< int >&& mt,
+              std::vector< double >&& pa, std::vector< double >&& pb ) {
+
+  unsigned int npp = ma.size();
+  if ( ( mb.size() != npp ) or ( za.size() != npp ) or ( zb.size() != npp ) or
```
jlconlin: Ah, here is where you check that the sizes are correct.

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceChannels.hpp
```diff
@@ -0,0 +1,157 @@
...
+  *  @brief Return the number of channels
+   */
+  long numberChannels() const { return this->NCH(); }
+
+  /**
+   *  @brief Return the particle pair numbers of each channel
+   */
+  auto PPI() const {
+    return ListRecord::list()
+             | ranges::view::stride( 6 )
```
jlconlin: Similar comment as before; does it make sense to write a private function to handle these?

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceChannels/src/generateList.hpp
```diff
@@ -0,0 +1,33 @@
+static std::vector< double >
+generateList( std::vector< double >&& l,
```
jlconlin: Does it make sense to put a check like this on the `ListRecord`? Then we could just catch an exception and add additional information before re-throwing the exception.

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceParameters.hpp
```diff
@@ -0,0 +1,71 @@
...
+  /**
+  *  @brief Return the resonance energies
+   */
+  auto resonanceEnergies() const { return this->ER(); }
+
+  /**
+   *  @brief Return the resonance parameters
+   */
+  auto GAM() const {
+    return ListRecord::list()
```
jlconlin: 👍

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceParameters/test/ResonanceParameters.test.cpp
```diff
@@ -0,0 +1,188 @@
...
+    " 0.000000+0 0.000000+0          0          2         24          42625 2151     \n"
+    "-1.223300+6 1.000000+0 9.611086+5 2.000000+0 3.000000+0 4.000000+02625 2151     \n"
+    " 5.000000+0 0.000000+0 0.000000+0 0.000000+0 0.000000+0 0.000000+02625 2151     \n"
+    " 7.788000+3 1.455000+0 1.187354+3 6.000000+0 7.000000+0 8.000000+02625 2151     \n"
+    " 9.000000+0 0.000000+0 0.000000+0 0.000000+0 0.000000+0 0.000000+02625 2151     \n";
+}
+
+void verifyChunk( const ResonanceParameters& chunk ) {
+
+  CHECK( 2 == chunk.NRS() );
```
jlconlin: So a dirty little secret. I don't double check these values as I review a Pull Request.

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/src/verifySize.hpp
```diff
@@ -0,0 +1,11 @@
+static void
+verifySize( int NJS ) {
+
+  if ( NJS == 0 ) {
```
jlconlin: Should this check `NJS < 1`, or `NJX <= 0`?

src/ENDFtk/resonanceParameters/resolved/BreitWignerReichMooreBase/src/verifySize.hpp
whaeck: The value we get from NLS() - something.size() - is essentially an unsigned int, even though it is converted to an int because of, you know, ENDF. As a result, this function will never be called with a negative value. If you only look at the function signature, then yes it should be.

src/ENDFtk/resonanceParameters/resolved/BreitWignerReichMooreLValueBase/src/generateList.hpp
whaeck: No, these are individual double values from each vector. It would copy them anyway.

src/ENDFtk/resonanceParameters/resolved/BreitWignerLValue/Resonance.hpp
whaeck: In the resonances() function on BreitWignerLValue.

src/ENDFtk/resonanceParameters/resolved/BreitWignerReichMooreLValueBase.hpp
whaeck: I initially did, but I'm getting 'invalid use of void expression' compiler errors. I now found out why, so I changed it.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBaseWithoutSpin/src/ctor.hpp
```diff
@@ -0,0 +1,12 @@
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] lvalues   the l values and the resonance parameters
+ */
+UnresolvedBaseWithoutSpin(
+  std::vector< LValue >&& lvalues ) :
+    // no need for a try ... catch: nothing can go wrong here
```
jlconlin: I understand what you are saying here, but these kinds of statements make me nervous.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBaseWithoutSpin/src/verifySize.hpp
```diff
@@ -0,0 +1,11 @@
+static void
+verifySize( int NLS ) {
+
+  if ( NLS == 0 ) {
```
jlconlin: `NLS <= 0`?

src/ENDFtk/resonanceParameters/unresolved/EnergyDependent.hpp
```diff
@@ -1,49 +1,66 @@
...
-  
-  double el;
-  double eh;
-  int naps;
-  double spi;
-  double ap;
-  int lssf;
+class EnergyDependent;
+template <>
+struct LJValueType< EnergyDependent > {
```
jlconlin: So, I'm not sure what this class provides. One could include lines 5–6 of directly instead of lines 27–28 and have the same functionality. Am I missing something?

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths.hpp
```diff
@@ -1,52 +1,104 @@
-class EnergyDependentFissionWidths {
+class EnergyDependentFissionWidths;
+template <>
+struct LJValueType< EnergyDependentFissionWidths > {
+
+  #include "ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths/JValue.hpp"
```
jlconlin: Again, why not put these include statements directly in the `EnergyDependentFissionWidths` class?

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths/JValue/src/generateList.hpp
```diff
@@ -0,0 +1,13 @@
+static std::vector< double >
+generateList( double spin,
```
jlconlin: I think their should be a generic `generateList` somewhere; perhaps as a factory function for the `ListRecord`. This function is virtually repeated in many different places.

```cpp
template< typename... Args >
std::vector< double >
generateList( Args&& args... ){

...
}
```

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths/JValue/src/verifySize.hpp
```diff
@@ -0,0 +1,9 @@
+static void
+verifySize( unsigned int NE ) {
+
+  if ( NE == 0 ) {
```
jlconlin: `NE <= 0`

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent.hpp
```diff
@@ -1,51 +1,66 @@
...
-  int naps;
-  double spi;
-  double ap;
-  int lssf;
-  
+class EnergyIndependent;
+template <>
+struct LJValueType< EnergyIndependent > {
+
+  #include "ENDFtk/resonanceParameters/unresolved/EnergyIndependent/JValue.hpp"
```
jlconlin: Again, should these `#include` statements be inside the `EnergyIndependent` class

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ParticlePairs.hpp
whaeck: No need, we test on size. The size is always a multiple of 12. We don't test on an empty list, I'm changing that.

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ParticlePairs/src/generateList.hpp
whaeck: Only for the user provided data.

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent/JValue.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+  /**
+   *  @brief Return the number of degrees of freedom for the neutron width.
+   */
+  int neutronWidthDegreesFreedom() const { return this->AMUN(); }
+
+  /**
+   *  @brief Return the number of degrees of freedom for the gamma width.
+   */
+  static constexpr int AMUG() { return 0; }
```
jlconlin: Is this function (and similar ones) just to provide a common interface? If so, why not put this function in the base class and override it in the daughter classes where it needs to be overridden?

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent/JValue.hpp
```diff
@@ -0,0 +1,126 @@
...
+ *
+ *  This JValue class applies to the case of energy independent widths (and
+ *  fission is absent).
+ *
+ *  See ENDF102, section 2.3.1 for more information.
+ */
+template < typename Range >
+class JValue {
+
+  Range chunk;
```
jlconlin: As before, what is `Range chunk`? Should we be more descriptive here?

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceChannels/src/generateList.hpp
whaeck: No, because each case where we need to flatten data is different. Particle pairs are in sets of 12 values, resonance channels in sets of 5 (with a sixth value equal to an index), etc.

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceParameters/test/ResonanceParameters.test.cpp
whaeck: You should look at the tests to see what is tested. The actual values do not really matter.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBase/src/NC.hpp
```diff
@@ -0,0 +1,9 @@
+/**
+ *  @brief Return the number of lines in this MF2 MT151 component
+ */
+long NC() const {
+
+  long NC = 1;
+  for ( const auto& lvalue : this->lValues() ) { NC += lvalue.NC(); }
```
jlconlin: I assume there is some sort of `ranges` magic to do this. I don't really care, but ranges is more fun to use.

src/ENDFtk/section/2/151/src/ctor.hpp
```diff
@@ -0,0 +1,56 @@
...
+ *  @brief Special case constructor (only scattering radius is given)
+ *
+ *  @param[in] zaid   the material ZAID value
+ *  @param[in] awr    the atomic weight ratio
+ *  @param[in] el     the lower limit of the resonance range
+ *  @param[in] eh     the upper limit of the resonance range
+ *  @param[in] spin   the spin of the target nucleus
+ *  @param[in] ap     the scattering radius (in units of 10^-12 cm)
+ */
+Type( double zaid, double awr, double el, double eh, double spin, double ap ) :
```
jlconlin: It doesn't look like this one can be called from the `Iterator` version of the constructor. Should it? My guess is probably not, but the question needs to be asked.

src/ENDFtk/section/2/151/src/readIsotopes.hpp
```diff
@@ -0,0 +1,19 @@
+template< typename Iterator >
+std::vector< resonanceParameters::Isotope >
+readIsotopes( const HEAD& head,
```
jlconlin: I don't think this will work when we have the special case. Do we have a test for a special case using this "sub-constructor"?

src/ENDFtk/section/2/151/src/verifyNIS.hpp
```diff
@@ -0,0 +1,10 @@
+static void
+verifyNIS( int NIS ) {
+
+  if ( NIS == 0 ) {
```
jlconlin: `NIS <= 0`

src/ENDFtk/section/2/151/src/verifyNIS.hpp
```diff
@@ -0,0 +1,10 @@
+static void
+verifyNIS( int NIS ) {
+
+  if ( NIS == 0 ) {
+
+    Log::error( "No isotopes with resonance parameters were defined" );
+    Log::info( "NIS must be non zero" );
```
jlconlin: `NIS must be greater than zero`

src/ENDFtk/section/2/151/test/2.test.cpp
```diff
@@ -57,14 +60,14 @@ SCENARIO( "section::Type<2>" ) {
...
-  GIVEN( "a valid instance of section::Type< 2 >" ) {
+  GIVEN( "a valid instance of section::Type< 2, 151 >" ) {
 
     std::string string = chunkSpecialCase() + validSEND();
     auto begin = string.begin();
     auto end = string.end();
     long lineNumber = 1;
     HeadRecord head( begin, end, lineNumber );
-    section::Type< 2 > chunk( head, begin, end, lineNumber, 125 );
+    section::Type< 2, 151 > chunk( head, begin, end, lineNumber, 125 );
```
jlconlin: This looks like where we have a `SpecialCase` that uses the Iterator constructor, right?

src/ENDFtk/section/2/151/test/2.test.cpp
```diff
@@ -134,16 +137,15 @@ SCENARIO( "section::Type<2>" ) {
...
       /*
+      THEN( "a section::Type< 2, 151 > can be constructed" ){
+        REQUIRE_THROWS( section2151( HEAD, begin, end, lineNumber, MAT ) );
         REQUIRE( 151 == MF2.MT() );
         REQUIRE( 17035 == MF2.ZA() );
         REQUIRE( 34.6685 == Approx( MF2.atomicWeightRatio() ) );
         REQUIRE( 1 == MF2.isotopes.size() );
-      */
       }
+      */
```
jlconlin: Should this be uncommented or deleted?

src/ENDFtk/section/BaseWithoutMT.hpp
```diff
@@ -6,19 +6,35 @@ class BaseWithoutMT {
...
-  
+
   /* constructor */
   #include "ENDFtk/section/BaseWithoutMT/src/ctor.hpp"
 
 public:
 
   /* get methods */
+
+  /**
```
jlconlin: You are the best about documentation.

src/ENDFtk/resonanceParameters/resolved/ReichMoore/test/ReichMoore.test.cpp
```diff
@@ -5,209 +5,260 @@
...
+      int nlsc = 1; // should be larger than or equal to 2
+      std::vector< LValue > lvalues =
+        { { 1.982069e+1, 0.0, 0,
+            { -1.470000e+5, 4.730000e+5 }, { 0.5, 0.5 },
+            { 5.430695e+2, 1.072906e+5 }, { 3.680695e+2, 1.072900e+5 },
+            { 1.750000e+2, 5.600000e-1 }, { 3., 4. } },
+          { 1.982069e+1, 0.0, 1,
+            { -2.060000e+0, 5.160000+0 }, { 0.5, 0.5 },
+            { 3.006610e-2, 3.393822e-2 }, { 5.750000e-3, 3.920000e-3 },
+            { 2.430000e-2, 3.000000e-2 }, { 1.610000e-5, 1.822000e-5 } } };
```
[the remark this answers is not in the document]
whaeck: The data that goes with the test is just in front of where it is being used to be consistent. How would putting this at the beginning of the test make this more readable?

src/ENDFtk/resonanceParameters/unresolved/EnergyDependent/JValue.hpp
```diff
@@ -0,0 +1,170 @@
...
+   *  @brief Return the number of interpolation ranges (always 1).
+   */
+  static constexpr int NR() { return 1; }
+
+  /**
+   *  @brief Return the interpolants
+   */
+  auto interpolants() const {
+
+    return std::array< int, 1 >{ this->INT() };
```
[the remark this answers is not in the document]
whaeck: interpolants() never returns a vector anywhere else. It is a ranges::view::all( ... ) of a vector.
This was added to make the JValue list record feel like a tab1.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBase.hpp
```diff
@@ -0,0 +1,53 @@
+/**
+ *  @class
+ *  @brief Base class for the unresolved resonance paraneters
+ *
+ *  See ENDF102, section 2.3.1 for more information.
+ */
+template < typename Derived,
+           typename LValue = typename LJValueType< Derived >::ltype >
```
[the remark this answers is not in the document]
whaeck: This is going to be long.

This is a CRTP for a Derived class containing a template for the LValue class that depends on the derived class. Since we're going to use the CRTP for defining the derived class, we cannot define Derived::LValue in the CRTP class declaration since the Derived type is incomplete when the compiler resolves the inheritance.

For example:
```cpp
template < typename Derived > class UnresolvedBase {

  std::vector< Derived::LValue > lvalues_;
}

class EnergyDependent : protected UnresolvedBase< EnergyDependent > {

  class LValue {};
}
``` 
When the compiler looks at UnresolvedBase< EnergyDependent >, EnergyDependent is incomplete. The compiler does not even know about the EnergyDependent::LValue class yet.

The templated struct is used to maintain a simple name for the LValue class since it has to be declared outside of its Derived class declaration (otherwise I would need a EnergyDependentLValue class). I also declare an alias to the LValue class inside the EnergyDependent class so that I can use EnergyDependent::LValue later on as if I had declared it inside the class. It's all about interface.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBaseWithoutSpin.hpp
```diff
@@ -0,0 +1,87 @@
...
+struct LJValueType {};
+
+/**
+ *  @class
+ *  @brief Base class for the unresolved resonance paraneters
+ *
+ *  See ENDF102, section 2.3.1 for more information.
+ */
+template < typename Derived,
+           typename LValue = typename LJValueType< Derived >::ltype >
```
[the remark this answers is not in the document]
whaeck: See above

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBaseWithoutSpin.hpp
```diff
@@ -0,0 +1,87 @@
...
+   */
+  int averageFissionWidthFlag() const {
+
+    return static_cast< const Derived* >( this )->LFW();
+  }
+
+  /**
+   *  @brief Return the target spin
+   */
+  double spin() const {
```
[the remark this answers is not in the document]
whaeck: There are three options for the unresolved resonances. Two of them use an initial CONT record that contains only ap, lssf and the spin value. The third option is a case where this information is on a LIST record that contains energy values for the energy dependent fission widths.

UnresolvedBaseWithoutSpin represents the base interface for all three options. Because this is a CRTP I can decide in the derived class how to handle spin, ap and lssf. UnresolvedBase is derived from UnresolvedBaseWithoutSpin and stores the ap, spin and lssf value inside it. EnergyDependent and EnergyIndependent then derive from UnresolvedBase (because they store this data in a CONT) while EnergyDependentFissionWidths derives from UnresolvedBaseWithoutSpin because I do not need to store the spin, ap and lssf value in it since I have it in the list with energy values (the ListRecord stored in there).

BTW: for the sections, we use a similar approach: Base and BaseWithoutMT. Enumerated sections like MF2 MT151 still have an MT number but we don't store it since it is 151.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBaseWithoutSpin/src/ctor.hpp
whaeck: That's a note for me so I can understand why there is no try ... catch when I come back to it. 
There are two reasons: nothing can go wrong so putting a try ...catch would lead to unreachable code, or it is handled in the derived class. In this case, the comment is the wrong one: it should be handled in the derived class.

src/ENDFtk/resonanceParameters/unresolved/EnergyDependent.hpp
whaeck: See above

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths.hpp
whaeck: See above

src/ENDFtk/resonanceParameters/unresolved/EnergyDependentFissionWidths/JValue/src/generateList.hpp
whaeck: Maybe, but I would still have to specialise something to do proper diagnostics and error messages.

Also: in some cases, I would like to append the different vectors, in other cases I would like to consider them columns and in other cases I need padding.

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent.hpp
whaeck: See above

src/ENDFtk/resonanceParameters/resolved/BreitWignerLValue/Resonance.hpp
jlconlin: Ok, I saw it later on. Thanks for pointing it out.

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent/JValue.hpp
whaeck: Yes, but the common interface for JValues would be limited to the degrees of freedom (AMU* variables) and the spin value. The widths could not be part of the common interface since they are only single values for energy independent widths and vectors for fully energy dependent widths. As such, the added value for a common base class is negligeable in my opinion.

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent/JValue.hpp
whaeck: It is a range created using chunk(). Can't be more descriptive than that :-D

This is actually an intermediate type that the user will not create directly. It is here for interface purposes.

By the way, that's what Austin used as its name in many other cases like this.

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/ResonanceParameters/test/ResonanceParameters.test.cpp
jlconlin: I do look at the tests, I just don't look closely at the values themselves.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBase/src/NC.hpp
whaeck: ranges::accumulate
But: really? It will provide no added value changing this. And it's done like this all over.

src/ENDFtk/section/2/151/src/ctor.hpp
whaeck: This is a convenience constructor to create the special case of one isotope, with only the channel radius defined (LRU=0). It is useful when creating an ENDF file that only needs this for MF2. PENDF files will use it for instance.

src/ENDFtk/section/2/151/src/readIsotopes.hpp
whaeck: It does handle the special case, in the readParameters() function called in the Isotope constructors.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBase.hpp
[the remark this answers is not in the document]
jlconlin: Good explanation.

src/ENDFtk/section/2/151/test/2.test.cpp
whaeck: yes, that's right.

src/ENDFtk/section/2/151/test/2.test.cpp
whaeck: I need a valid LRF7 chunk here. The one that was there before is an incomplete piece and makes the test fail. I'll fix those tests later.

src/ENDFtk/resonanceParameters/unresolved/EnergyIndependent/JValue.hpp
jlconlin: Just because Austin gave it a poor name doesn't mean we need to continue using a poor name.

src/ENDFtk/resonanceParameters/unresolved/UnresolvedBase/src/NC.hpp
jlconlin: This was not intended to be a serious comment.

jlconlin:
I hate to break it to you, but not everything compiles.

```console
[ 82%] Building CXX object src/ENDFtk/file/2/test/CMakeFiles/ENDFtk.file.2.test.dir/2.test.cpp.o
In file included from /Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/file/2/test/2.test.cpp:4:
In file included from /Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk.hpp:22:
In file included from /Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/hana-adapter/src/include/boost/hana.hpp:110:
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/hana-adapter/src/include/boost/hana/for_each.hpp:31:9: error: static_assert failed due to requirement
      'hana::Foldable<boost::hana::tuple<> ()>::value' "hana::for_each(xs, f) requires 'xs' to be Foldable"
        static_assert(hana::Foldable<S>::value,
        ^             ~~~~~~~~~~~~~~~~~~~~~~~~
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/file/2/src/print.hpp:4:3: note: in instantiation of function template specialization
      'boost::hana::for_each_t::operator()<boost::hana::tuple<> (&)(), (lambda at /Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/file/2/src/print.hpp:4:37)>' requested here
  hana::for_each( optionalSections, [&]( auto MT ){
  ^
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/file/2/test/2.test.cpp:179:13: note: in instantiation of function template specialization
      'njoy::ENDFtk::file::Type<2>::print<std::__1::back_insert_iterator<std::__1::basic_string<char> > >' requested here
      file2.print( output, 1122 );
            ^
In file included from /Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/file/2/test/2.test.cpp:4:
In file included from /Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk.hpp:22:
In file included from /Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/hana-adapter/src/include/boost/hana.hpp:110:
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/hana-adapter/src/include/boost/hana/for_each.hpp:35:25: error: call to deleted member function 'apply'
        return ForEach::apply(static_cast<Xs&&>(xs), static_cast<F&&>(f));
               ~~~~~~~~~^~~~~
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/hana-adapter/src/include/boost/hana/detail/dispatch_if.hpp:21:31: note: candidate function [with T = <boost::hana::tuple<> (&)(),
      (lambda at /Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/file/2/src/print.hpp:4:37)>] has been explicitly deleted
        static constexpr auto apply(T&& ...) = delete;
                              ^
2 errors generated.
```

whaeck:
@jlconlin I knew there was a reason that code was commented out in print() or that other function. Just a sec ...

whaeck:
@jlconlin It was as I thought. You cannot use a hana tuple that is empty. It works again.

It would have surprised me that there was an issue there. That code is very similar to file 7 and 8 and those work properly.

jlconlin:
Sorry, still not compiling. 

```
[ 87%] Building CXX object src/ENDFtk/section/2/151/test/CMakeFiles/ENDFtk.section.2.151.test.dir/2.test.cpp.o
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/section/2/151/test/2.test.cpp:58:29: error: reference to non-static member function must be called; did you mean to call it with no
      arguments?
        REQUIRE( 1 == chunk.isotopes.size() );
                      ~~~~~~^~~~~~~~
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/catch-adapter/src/single_include/catch.hpp:11594:91: note: expanded from macro 'REQUIRE'
#define REQUIRE( expr ) INTERNAL_CATCH_TEST( "REQUIRE", Catch::ResultDisposition::Normal, expr  )
                                                                                          ^~~~
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/catch-adapter/src/single_include/catch.hpp:2229:32: note: expanded from macro 'INTERNAL_CATCH_TEST'
            ( __catchResult <= expr ).endExpression(); \
                               ^~~~
/Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/section/2/151/test/2.test.cpp:58:29: error: reference to non-static member function must be called; did you mean to call it with no
      arguments?
        REQUIRE( 1 == chunk.isotopes.size() );
                      ~~~~~~^~~~~~~~
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/catch-adapter/src/single_include/catch.hpp:11594:91: note: expanded from macro 'REQUIRE'
#define REQUIRE( expr ) INTERNAL_CATCH_TEST( "REQUIRE", Catch::ResultDisposition::Normal, expr  )
                                                                                          ^~~~
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/catch-adapter/src/single_include/catch.hpp:2236:60: note: expanded from macro 'INTERNAL_CATCH_TEST'
    } while( Catch::isTrue( false && static_cast<bool>( !!(expr) ) ) ) // expr here is never evaluated at runtime but it forces the compiler to give it a look

```
Did you compile this?

whaeck:
@jlconlin Yes, before I changed it :-)

jlconlin:
Well, everything compiled, but several tests failed:

```
96% tests passed, 5 tests failed out of 126

Total Test time (real) =  42.41 sec

The following tests FAILED:
         10 - ENDFtk.resonanceParameters.resolved.RMatrixLimited.ResonanceParameters (SEGFAULT)
         25 - ENDFtk.resonanceParameters.resolved.RMatrixLimited.ResonanceChannels (Failed)
         52 - ENDFtk.resonanceParameters.resolved.RMatrixLimited (Failed)
         58 - ENDFtk.resonanceParameters.resolved.RMatrixLimited.ParticlePairs (Failed)
         89 - ENDFtk.resonanceParameters.resolved.RMatrixLimited.SpinGroup (Failed)
Errors while running CTest
```

whaeck:
@jlconlin All tests compile and run successfully.

jlconlin:
I'm still getting a compiler error using GCC 9.3.0 on my Mac:

```console
[ 97%] Building CXX object src/ENDFtk/resonanceParameters/resolved/ReichMooreLValue/Resonance/test/CMakeFiles/ENDFtk.resonanceParameters.resolved.ReichMooreLValue.Resonance.test.dir/Resonance.test.cpp.o
In file included from /Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/utility/iterator_concepts.hpp:23,
                 from /Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/begin_end.hpp:26,
                 from /Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/core.hpp:17,
                 from /Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/all.hpp:17,
                 from /Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk.hpp:19,
                 from /Users/jlconlin/NJOY21/Code/ENDFtk/src/ENDFtk/resonanceParameters/resolved/ReichMooreLValue/Resonance/test/Resonance.test.cpp:4:
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/utility/functional.hpp: In instantiation of 'ranges::v3::composed<First, Second>::composed(Second, First) [with Second = ranges::v3::detail::emplace_fn<ranges::v3::variant<void, void>, 0>; First = ranges::v3::detail::unique_visitor<ranges::v3::variant<const double&>, ranges::v3::variant<const double&, const double&> >]':
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/utility/functional.hpp:437:60:   required from 'ranges::v3::composed<Second, First> ranges::v3::compose_fn::operator()(Second, First) const [with Second = ranges::v3::detail::emplace_fn<ranges::v3::variant<void, void>, 0>; First = ranges::v3::detail::unique_visitor<ranges::v3::variant<const double&>, ranges::v3::variant<const double&, const double&> >]'
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/detail/variant.hpp:453:17:   required from 'decltype (ranges::v3::compose(ranges::v3::detail::emplace_fn<Variant, N>{((ranges::v3::detail::variant_visitor<Fun, Variant>*)this)->ranges::v3::detail::variant_visitor<Fun, Variant>::var_}, ((ranges::v3::detail::variant_visitor<Fun, Variant>*)this)->ranges::v3::detail::variant_visitor<Fun, Variant>::fun_)(u)) ranges::v3::detail::variant_visitor<Fun, Variant>::operator()(ranges::v3::indexed_element<U, M>) [with U = const double&; long unsigned int N = 0; Fun = ranges::v3::detail::unique_visitor<ranges::v3::variant<const double&>, ranges::v3::variant<const double&, const double&> >; Variant = ranges::v3::variant<void, void>; decltype (ranges::v3::compose(ranges::v3::detail::emplace_fn<Variant, N>{((ranges::v3::detail::variant_visitor<Fun, Variant>*)this)->ranges::v3::detail::variant_visitor<Fun, Variant>::var_}, ((ranges::v3::detail::variant_visitor<Fun, Variant>*)this)->ranges::v3::detail::variant_visitor<Fun, Variant>::fun_)(u)) = void]'
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/utility/invoke.hpp:124:13:   required from 'constexpr decltype (static_cast<F&&>(fn)(static_cast<Args&>(ranges::v3::invoke_fn::operator()::args) ...)) ranges::v3::invoke_fn::operator()(F&&, Args&& ...) const [with F = ranges::v3::detail::variant_visitor<ranges::v3::detail::unique_visitor<ranges::v3::variant<const double&>, ranges::v3::variant<const double&, const double&> >, ranges::v3::variant<void, void> >&; Args = {ranges::v3::indexed_element<const double&, 0>}; typename meta::detail::_if_<meta::list<std::integral_constant<bool, (! std::is_member_pointer<typename std::remove_cv<typename std::remove_reference<_Tp>::type>::type>::value)>, int>, std::integral_constant<bool, true> >::type <anonymous> = 0; decltype (static_cast<F&&>(fn)(static_cast<Args&>(ranges::v3::invoke_fn::operator()::args) ...)) = void]'
/Users/jlconlin/NJOY21/Code/ENDFtk/subprojects/range-v3-adapter/src/include/range/v3/detail/variant.hpp:311:47:   required from 'constexpr int ranges::v3::detail::variant_visit_(std:
```

Of course this goes on forever.

whaeck:
Looks like a std::variant or range issue. This compiles using gcc/7.4.0 on my personal computer. I'll see if I can compile with gcc/9.x to see what the actual problem is.

src/ENDFtk/resonanceParameters/resolved/BreitWignerReichMooreLValueBase.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+protected:
+
+  /* constructor */
+  #include "ENDFtk/resonanceParameters/resolved/BreitWignerReichMooreLValueBase/src/ctor.hpp"
+
+  /**
+   *  @brief Return the first width
+   */
+  auto G1() const { return BreitWignerReichMooreLValueBase::column( 2 ); }
```
jlconlin: 👍

src/ENDFtk/section/2/151/test/2.test.cpp
```diff
@@ -134,16 +137,15 @@ SCENARIO( "section::Type<2>" ) {
...
+      THEN( "a section::Type< 2, 151 > can be constructed" ){
+        REQUIRE_THROWS( section2151( HEAD, begin, end, lineNumber, MAT ) );
         REQUIRE( 151 == MF2.MT() );
         REQUIRE( 17035 == MF2.ZA() );
         REQUIRE( 34.6685 == Approx( MF2.atomicWeightRatio() ) );
-        REQUIRE( 1 == MF2.isotopes.size() );
-      */
+        REQUIRE( 1 == MF2.isotopes().size() );
       }
+      */
```
jlconlin: This still needs to be deleted.

whaeck: That's done now

whaeck:
@jlconlin We're ready to go if you are ;-)

Next up resonanceReconstruction update to this interface.

PR #68, opened 2020-04-09
Moving function to better align with Style Guide.

jlconlin:
I just moved `makeTape` from the [Tape header](https://github.com/njoy/ENDFtk/blob/master/src/ENDFtk/syntaxTree/Tape.hpp#L50) to it's own file, more in alignment with the [Style Guide](http://www.njoy21.io/DevelopersGuide/Style/)

I was looking for where this function was defined and struggled to find it because it wasn't in it's own file.


jlconlin:
This should be really short and sweet.

jlconlin:
This also takes care of #67

PR #69, opened 2020-04-17
Because assignment and equality is not the same

whaeck:


jlconlin:
All tests passed. (Whether or not everything is tested, is quite another story.)

PR #73, opened 2020-06-03
Adding GENDF Capabilities to ENDFtk

nathangibson14:
This pull request adds the ability to read a GENDF file via the syntaxTree and to do basic parsing of cross section (MFD 3) and transfer matrix (MFD 6) sections.  This only supports reading GENDF files; writing is not yet implemented.

The classes of the syntaxTree are now tagged as ENDF or GENDF, and ENDF is default.  This means there's no change to the existing syntax for reading ENDF files.  Rather than specify the tag directly, users (and other places in the code) can use convenience aliases such as `GendfTape`.

Parsing is currently limited to extracting a cross section or matrix element by group, Legendre order, and dilution index.  Adding capability for an array of cross sections will likely follow sooner than later.

There is plenty more to do for full GENDF support, but I believe this is enough to start using while other functionality is developed.  Things to include later include:

- Support for header data (MFD 1) and delayed neutron data (MFD 5)
- Support for ratio quantities (MFD/MT 3/452 and various others)
- Support for compressed fission transfer matrix format (prod/spec for lower energy groups)
- Support for photon production
- More convenient data structures from parsed data structures.

The test suite passes, although there are plenty of `ranges`-related test suite quirks on various systems/compilers.  None of the quirks touch pieces of the code affected by this PR.  I found issues with two tests on Snow/GCC 8.3.0 and one test on my MacBook/GCC 9.2.0.  I'll put up a separate issue, but it's time to consider upgrading the ranges package for a variety of reasons.

Let me know how my biggest foray into C++ looks! :smile:


src/ENDFtk/section/Gendf3.hpp
```diff
@@ -0,0 +1,29 @@
+template<>
+class GendfType< 3 > : protected GendfBase {
+
+protected:
```
jlconlin, line 4: This doesn't need to be here with the way the class is currently configured.

src/ENDFtk/section/Gendf3/src/ctor.hpp
```diff
@@ -0,0 +1,15 @@
...
+GendfType( HeadRecord& head,
+           BufferIterator& begin,
+           const BufferIterator& end,
+           long& lineNumber,
+           int MAT )
+    : GendfBase( head, begin, end, lineNumber, MAT ) {
+
+  // TODO: Add support for ratio quantities
+  if( data_.at(num_groups_).NG2() == 3 ) {
+    Log::info( "ENDFtk does not yet support ratio quantities." );
```
jlconlin, line 11: If you are going to throw an exception, Log an error. You can provide informational logs afterwards as needed.

The practice we have developed is we log an error when it is initially thrown. If we catch (and possibly re-throw) an exception, then we log an informational message.

src/ENDFtk/section/Gendf3/src/getCrossSection.hpp
```diff
@@ -0,0 +1,44 @@
+/*
+ * @brief Return cross section value.
+ * @param group GENDF group index (1-based, low group = low energy)
+ * @param order Legendre order
+ * @param idil  Index of dilution (0-based)
+ */
+double getCrossSection( int group,
```
jlconlin, line 7: Ugh. I hate getters (or setters) that start like `get...` (or `set...`). The `get` is unnecessary. You can tell the difference between a getter and a setter based on the function arguments. Granted, this is a matter of opinion (mine, in fact), but it is what we have adopted for NJOY. See Section 6 of our [Style Guide](http://www.njoy21.io/DevelopersGuide/Style/Code.html#Class).

src/ENDFtk/section/Gendf3/test/Gendf3.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "catch.hpp"
+#include "ENDFtk.hpp"
+
+using namespace njoy::ENDFtk;
+
+
+std::string getSection();
+
+
+SCENARIO( "creating a GendfBase object" ) {
```
jlconlin, line 12: You are not creating a GendfBase object here. You are creating a GendfType<3>. Yeah, I'm being pedantic. I don't really care what is said here, but we should try to be accurate.

src/ENDFtk/section/Gendf3/test/Gendf3.test.cpp
```diff
@@ -0,0 +1,84 @@
...
+      std::string buffer = getSection();
+
+      auto begin = buffer.begin();
+      auto end = buffer.end();
+      long lineNo = 0;
+      int MAT = 9237;
+      HeadRecord head(begin, end, lineNo);
+
+      THEN( "the object can be created" ) {
+        section::GendfType<3> chunk(head, begin, end, lineNo, MAT );
```
jlconlin, line 25: `chunk`? I see you've taken after @whaeck  in your naming convention

src/ENDFtk/GendfDataRecord.hpp
```diff
@@ -0,0 +1,46 @@
+class GendfDataRecord : protected ListRecord {
```
jlconlin, line 1: I think I'd prefer `GENDFDataRecord` instead of `GendfDataRecord`. I'm not sure it warrants changing everything, though.

src/ENDFtk/section/Gendf6/src/ctor.hpp
```diff
@@ -0,0 +1,16 @@
...
+GendfType( HeadRecord& head,
+           BufferIterator& begin,
+           const BufferIterator& end,
+           long& lineNumber,
+           int MAT )
+    : GendfBase( head, begin, end, lineNumber, MAT ) {
+
+  // TODO: Add support for compressed fission format
+  if( data_.count(0) ) {
+    Log::info( "ENDFtk does not yet have support "
```
jlconlin, line 11: Same comment as before about logging an error.

src/ENDFtk/section/Gendf6/src/getMatrixElement.hpp
```diff
@@ -0,0 +1,66 @@
+/*
+ * @brief Return matrix element from transfer matrix.
+ * @param in_group  Incoming energy group (GENDF ordering)
+ * @param out_group Outgoing energy group (GENDF ordering)
+ * @param order     Legendre order
+ * @param idil      Index of dilution (0-based)
+ */
+double getMatrixElement( int in_group,
```
jlconlin, line 8: Same comment as before about `get`ers

src/ENDFtk/section/Gendf6/src/getMatrixElement.hpp
```diff
@@ -0,0 +1,66 @@
...
+                         int out_group,
+                         int order,
+                         int idil ) {
+
+  // error checking
+  if( order > legendreOrder() ) {
+    Log::info( "{} is greater than Legendre order of dataset ({})",
```
jlconlin, line 15: `Log::error` here and elsewhere in this file.

src/ENDFtk/section/Gendf6/test/Gendf6.test.cpp
```diff
@@ -0,0 +1,168 @@
...
+        CHECK( chunk.getMatrixElement(4, 3, 2, 1) == Approx(-1.461486e-6) );
+        CHECK( chunk.getMatrixElement(4, 1, 0, 0) == 0.0 );
+        CHECK_THROWS( chunk.getMatrixElement(1, 1, 3, 0) );
+        CHECK_THROWS( chunk.getMatrixElement(1, 1, 0, 3) );
+        CHECK_THROWS( chunk.getMatrixElement(1, 0, 0, 0) );
+        CHECK_THROWS( chunk.getMatrixElement(0, 1, 0, 0) );
+      }
+    }
+  }
+}
```
jlconlin, line 124: You didn't test to make sure exceptions are thrown with bad data.

src/ENDFtk/section/GendfBase/src/ctor.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+  // find temperature and check all temperatures are consistent
+  bool temp_set = false;
+  for( const auto& [group, record] : data_ ) {
+    if( not temp_set ) {
+      temperature_ = record.temperature();
+      temp_set = true;
+    }
+    else if( record.temperature() != temperature_ ) {
+      Log::info( "Inconsistent temperature in MFD/MT {}/{}: {} != {}",
```
jlconlin, line 24: `Log::error`

src/ENDFtk/section/GendfBase/src/ctor.hpp
```diff
@@ -0,0 +1,30 @@
...
+           long& lineNumber,
+           int MAT )
+  : Base( head, MAT, head.MF() ),
+    num_legendre_( head.L1() ),
+    num_sigma0_( head.L2() ),
+    lrflag_( head.N1() ),
+    num_groups_( head.N2() ) {
+
+  // populate data records
+  data_ = populateData(head, begin, end, lineNumber, MAT);
```
jlconlin, line 14: `data_` should be instantiated in the block before the constructor begins.

```suggestion
  : Base( head, MAT, head.MF() ),
    data_(  populateData(head, begin, end, lineNumber, MAT) ),
    num_legendre_( head.L1() ),
    num_sigma0_( head.L2() ),
    lrflag_( head.N1() ),
    num_groups_( head.N2() ) {
```

src/ENDFtk/section/GendfBase/src/populateData.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+    // create next record
+    next_record = GendfDataRecord( begin, end, lineNumber,
+                                   num_legendre_, num_sigma0_,
+                                   MAT, head.MF(), head.MT() );
+    group = next_record.group();
+
+    // check if group was already seen
+    if (data_records.count( group ) > 0) {
+      Log::info( "Group {} appears multiple times in MFD/MT {}/{}",
```
jlconlin, line 29: `Log::error`

src/ENDFtk/section/GendfBase/test/GendfBase.test.cpp
```diff
@@ -0,0 +1,162 @@
...
+using namespace njoy::ENDFtk;
+
+
+std::string validFEND();
+std::string mfd3mt2();
+std::string validSEND();
+std::string invalidSEND();
+
+
+SCENARIO( "creating a GendfBase object" ) {
```
jlconlin, line 15: Here you are testing a GendfBase.

src/ENDFtk/syntaxTree/Section.hpp
```diff
@@ -21,3 +23,54 @@ class Section {
...
+public:
+
+  /* ctor */
+  using Base = SectionBase< BufferIterator >;
+  using Base::SectionBase;
+
+  /* methods */
+
+  // this file implements the parse() function
+  #include "ENDFtk/syntaxTree/Section/src/parse-gendf.hpp"
```
jlconlin, line 62: I wonder if this could be implemented with a templated `parse` function:

```cpp
template< typename Tag >
void parse< Tag >(...);

template<>
void parse< GENDFTag >(...){}

template<>
void parse< ENDFTag >(...)
```

This way, you don't need to specialize the class, just the parse function. It seems that everything else is the same.

src/ENDFtk/syntaxTree/Section/test/GendfSection.test.cpp
```diff
@@ -0,0 +1,201 @@
...
+      AND_THEN( "the section can be parsed" ){
+        auto parsec = sectionTree.parse<3>();
+        REQUIRE( parsec.numGroups() == 44 );
+        REQUIRE( parsec.getCrossSection(1) == Approx(14.69141) );
+      }
+
+
+    } // WHEN
+
+    WHEN( "an invalid (MT!=0) SEND record ends the Section" ){
```
jlconlin, line 70: Good job testing failure conditions. Are there others that also need to be tested?

src/ENDFtk/syntaxTree/GendfMaterial.hpp
```diff
@@ -0,0 +1,71 @@
+/* Template specialization for GendfMaterial */
+template< typename BufferIterator >
+class Material< BufferIterator, GENDFTag > {
```
jlconlin, line 3: I'm not sure I see the reason for a separate GENDF Material. (Granted I didn't look real close.) It seems it is virtually identical to the Material class, just uses the GENDFTag. What exactly is the difference and can we do template specialization on a member function instead of the entire class?

nathangibson14: There's a subtle but big difference between a `GendfMaterial` and a `Material`.  A `Material` is processed by looping over the `File`s it contains, creating a map of these `File` objects.  In a GENDF file, there is no guarantee that all sections with the same MFD are consecutive.  And to make matters worse, there's no guarantee that a MFD/MT section is unique (although I'd hope the repeated section is at least identical!).

So I implemented this so that there is no `GendfFile`, and the `GendfMaterial` contains a map of `GendfSection` objects directly.

There are lots of places where the ENDF format is frustrating, but GENDF is much more frustrating despite having a much smaller spec!

src/ENDFtk/syntaxTree/Section.hpp
nathangibson14: I really wanted to do this, but I couldn't figure out a way to make it happen.  A tag with a default value doesn't play nice with the `int... OptionalMT` that `parse` is already templated on.  I also don't want the user to have to specify the tag.

If you can see a way to clean this up, I'm all for it.

src/ENDFtk/section/GendfBase/src/ctor.hpp
nathangibson14: This was an arguably sloppy way of dealing with an issue I had.  `populateData` needs access to `num_legendre_` and others.  I used the member variables inside `populateData`, which means I needed them to be set beforehand.  I could use `head.L1()` instead.  Or I could do the cascading constructor trick that @whaeck is a fan of.  I'll try to see what feels the most natural, as I think all the solutions are about as sloppy as the one I went with originally.

src/ENDFtk/GendfDataRecord.hpp
nathangibson14: I've debated acronyms in CamelCase for my entire coding career and have never definitively decided which is better.  Note that I didn't even stay consistent, as I used `GENDFTag` but `GendfType`.

It's pretty trivial to slap a `/Gendf/GENDF/` sed command onto these files, so if you feel more settled in this everlasting debate I have with myself.

src/ENDFtk/section/Gendf3/test/Gendf3.test.cpp
nathangibson14: I originally had `section`, but felt dirty using `section` as both a namespace and a variable at the same time.  And when I looked to other tests for guidance, voila -- I found @whaeck's approach.

whaeck:
Stop poking me ;-)

src/ENDFtk/section/Gendf3/test/Gendf3.test.cpp
nathangibson14: Copy and paste can be cruel.  I'll fix it.

src/ENDFtk/section/Gendf3/test/Gendf3.test.cpp
whaeck: Because that's what it is: a chunk of ENDF data :-p

src/ENDFtk/section/Gendf3/src/getCrossSection.hpp
nathangibson14: I like Python where I don't need `()` on my getters and setters :smile:

I usually am with you.  This one was because I was thinking of `getCrossSectionByGroup` versus `getCrossSection` returning the whole array.  I'm pretty sure I wrote those routines into OpenMOC years ago, so they came back to me.  I didn't even stick with that naming convention, so I won't argue in favor for `get` here!

src/ENDFtk/section/Gendf3/src/ctor.hpp
nathangibson14: I'm not convinced this is done consistently throughout, but I'm happy to change mine.  What you are suggesting makes more sense to me, I was just trying to stick to the convention I saw in similar places, which likely was non-conforming.

nathangibson14:
@whaeck Poke.

src/ENDFtk/section/Gendf3/test/Gendf3.test.cpp
nathangibson14: In my Python bindings, I've been using `parsec` for parsed sections.  I'm sure our friends in astrophysics would approve of that.

src/ENDFtk/section/Gendf3/src/ctor.hpp
whaeck: The way we've done it up to now is that an error is logged when the original exception is created, logging information later on to indicate where it happened, etc. by catching the exception.

src/ENDFtk/syntaxTree/Section.hpp
nathangibson14: In talking with Wim this morning, I think I found a cleaner solution in line with what you were saying.  Stay tuned!

src/ENDFtk/GendfDataRecord/test/GendfDataRecord.test.cpp
```diff
@@ -0,0 +1,46 @@
...
+    WHEN( "a valid record is given" ) {
+
+      std::string buffer = getMFD3Record();
+
+      auto begin = buffer.begin();
+      auto end = buffer.end();
+      long int lineNo = 2;
+
+      THEN( "the object can be created" ) {
+        GendfDataRecord record( begin, end, lineNo, 1, 1, 9228, 3, 2 );
```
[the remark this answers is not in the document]
nathangibson14: This is really just a wrapped `ListRecord`.  I think most of the error handling is covered there.

PR #74, opened 2020-06-03
Feature/mf2

whaeck:
Added a function to return SPI and AP for LRF7, to be similar to what other resonance formalisms return.

Beware: while SPI will return the spin of the target, AP will only return the first true scattering radius value it finds in the first spin group. If other spin groups use different radii, AP is meaningless and should not be used as a shortcut to retrieve the scattering radius.

### DIFF
--- a/src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/src/AP.hpp
+++ b/src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/src/AP.hpp
@@ -1,5 +1,10 @@
 /**
- *  @brief Return the target spin
+ *  @brief Return the scattering radius
+ *
+ *  This function will only return the first true scattering radius value it
+ *  finds in the first spin group. If other spin groups use different radii,
+ *  AP is meaningless and should not be used as a shortcut to retrieve
+ *  the scattering radius.
  */
 double AP() const {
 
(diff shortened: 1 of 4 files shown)


src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/src/SPI.hpp
```diff
@@ -0,0 +1,10 @@
+/**
+ *  @brief Return the target spin
+ */
+double SPI() const {
+
+  auto mt = this->particlePairs().MT();
+  unsigned int index = ranges::distance( ranges::begin( mt ),
+                                         ranges::find( mt, int( 2 ) ) );
```
jlconlin, line 8: I didn't know you could do this. This is great!

whaeck:
@jlconlin I changed some of the documentation

src/ENDFtk/resonanceParameters/resolved/RMatrixLimited/src/AP.hpp
```diff
@@ -0,0 +1,20 @@
+/**
+ *  @brief Return the scattering radius
+ *
+ *  This function will only return the first true scattering radius value it
+ *  finds in the first spin group. If other spin groups use different radii,
+ *  AP is meaningless and should not be used as a shortcut to retrieve
+ *  the scattering radius.
```
nathangibson14, line 7: Is a note saying "you shouldn't do this" a strong enough check?  Is it feasible to throw an exception if AP is called in a case where it's meaningless?

whaeck: True, but I don't  want to throw an exception at this point. LRF7 is not widely used and I think most evaluations use the same AP.

PR #82, opened 2020-09-04
Adding function to return list (vector) of MT numbers in a syntaxTree::File.

jlconlin:
This makes it possible to get all the MT numbers in the File without the client having to iterate through everything.

### DIFF
--- a/src/ENDFtk/syntaxTree/File.hpp
+++ b/src/ENDFtk/syntaxTree/File.hpp
@@ -53,6 +53,5 @@ public:
 
   int fileNumber() const { return this->MF(); }
 
-  std::vector< int > sectionNumbers() const { 
-    return ranges::view::keys( sections_ ); }
+  auto sectionNumbers() const { return ranges::view::keys( sections_ ); }
 };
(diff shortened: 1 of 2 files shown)


src/ENDFtk/syntaxTree/File.hpp
```diff
@@ -52,4 +52,7 @@ class File {
   int MF() const { return this->fileNo; }
 
   int fileNumber() const { return this->MF(); }
+
+  std::vector< int > sectionNumbers() const {
```
apmccartney, line 56: Now why would you take a nice lazy view and force the client to heap allocate memory for a container?

```suggestion
  auto sectionNumbers() const { 
```

src/ENDFtk/syntaxTree/File/test/File.test.cpp
```diff
@@ -29,10 +29,16 @@ SCENARIO( "Creating a syntax tree of an ENDF File" ){
           < syntaxTree::File< std::string::iterator > >
           ( head, start, beg, end, lin );
         
+        THEN( "We can exstract the MT numbers" ){
+          std::vector< int > refMTs{ 1, 2, 102 };
+          CHECK( refMTs == original->sectionNumbers() );
```
apmccartney, line 34: related to previous comment

```suggestion
          CHECK( ranges::equal( refMTs,original->sectionNumbers() ) );
```

jlconlin:
Thanks @apmccartney for the comments. (Glad to see you are still alive!) I have addressed them in my latest changes.

whaeck (approved):
I agree with @apmccartney comments so now we're good to go.

PR #83, opened 2020-09-17
Feature/lang3

whaeck:
Added a component to treat MF6 LAW=1 LANG=3, an internal NJOY ENDF component. It currently answers requirements for THERMR.

In addition to this new component, the following simplifications were also introduced:
- InterpolationSequenceRecord (a generic TAB2 followed by a sequence of stuff) has been revamped to allow for the use of simple ENDF components or std::variants of multiple ENDF. As a result, the read constructor originally in InterpolationSequenceRecord has been removed.
- All instances of SubSection composed of a std::variant of ENDF components has been removed to use the std::variant instead. This simplification will be beneficial for the python bindings.
- The introduction of a generic readSequence function to remove over 10 to 20 instances of similar functions on specific ENDF components.
- Added documentation in some MF6 components with heavy changes
- Refactored some tests to be like the others as well.

This pull request actually removes more code than it adds ;-)


src/ENDFtk/InterpolationSequenceRecord/src/ctor.hpp
```diff
@@ -1,38 +1,14 @@
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] interpolation   the interpolation record associated to the data
+ *  @param[in] sequence        the data
+ */
 InterpolationSequenceRecord ( InterpolationRecord&& interpolation,
-                              std::vector< Record >&& sequence ) :
+                              std::vector< Component >&& sequence ) :
```
jlconlin, line 8: Where is this `Component` defined?

src/ENDFtk/InterpolationSequenceRecord.hpp
```diff
@@ -1,27 +1,124 @@
...
+ *  @tparam Component        an ENDF component
+ *
+ *  @param[in] component     the ENDF component to be printed
+ *  @param[in] it            the current position in the output
+ *  @param[in] MAT           the MAT number
+ *  @param[in] MF            the MF number
+ *  @param[in] MT            the MT number
+ */
+template < typename OutputIterator, typename Component >
+void print( const Component& component,
```
jlconlin, line 16: I like these functions. They are very useful as can be seen in this Pull Request.

Why are these functions in this file?

src/ENDFtk/section/6/ChargedParticleElasticScattering.hpp
```diff
@@ -1,33 +1,96 @@
...
 class ChargedParticleElasticScattering {
+
 public:
 
   #include "ENDFtk/section/6/ChargedParticleElasticScattering/NuclearAmplitudeExpansion.hpp"
   #include "ENDFtk/section/6/ChargedParticleElasticScattering/LegendreCoefficients.hpp"
   #include "ENDFtk/section/6/ChargedParticleElasticScattering/NuclearPlusInterference.hpp"
-  #include "ENDFtk/section/6/ChargedParticleElasticScattering/SubSection.hpp"
+
+  using Variant = std::variant< NuclearAmplitudeExpansion, // LTP=1
```
jlconlin, line 19: 👍

jlconlin (commented):
Overall, I like how this removes a lot of code and reuses it. I don't quite follow everything, but I think I'll get it with a simple explanation.

jlconlin:
I'd also like some feedback from @ameliajo if this addresses her needs.

src/ENDFtk/InterpolationSequenceRecord/src/ctor.hpp
whaeck: That's a template parameter on the InterpolationSequenceRecord (should be in the hpp file). It is either a pure ENDF component or a variant of pure ENDF components.

src/ENDFtk/InterpolationSequenceRecord.hpp
whaeck: Since I want this to work on both a pure ENDF component and an std::variant of pure ENDF components, I need to define the print() and NC() function for both cases. They are in turn used inside the implementation of print() and NC() in the InterpolationSequenceRecord.

src/ENDFtk/InterpolationSequenceRecord/src/ctor.hpp
jlconlin: Oh, I didn't notice this was inside of a class. My bad.

src/ENDFtk/InterpolationSequenceRecord.hpp
jlconlin: I understand why you wrote them, but shouldn't these sit inside of `src/ENDFtk/src`?

jlconlin (approved):

src/ENDFtk/InterpolationSequenceRecord.hpp
whaeck: Since they are used with the InterpolationSequenceRecord, I would keep them with this component.

PR #85, opened 2020-09-27
Removed restrictions on LASYM values (NJOY uses more than just 0 and 1)

whaeck:


src/ENDFtk/section/7/4/src/ctor.hpp
```diff
@@ -27,11 +27,10 @@ Type( double zaid, double awr, int lat, int lasym,
   verifySecondaryTemperatures( this->b_.analyticalFunctionTypes(),
                                this->secondary_ );
   checkValue( this->LAT(), "LAT" );
-  checkValue( this->LASYM(), "LASYM" );
```
nathangibson14, line 30: Why are you getting rid of the check rather than fixing the check?

nathangibson14 (commented):
Hard to find the real changes in the sea of whitespace tweaks. :)

Looks fine to me overall, just left one question.

src/ENDFtk/section/7/4/src/ctor.hpp
whaeck: I don't know which values are being used internally by NJOY.

whaeck:
@nathangibson14 That's my editor, it automatically removed trailing white space.

src/ENDFtk/section/7/4/src/ctor.hpp
whaeck: I can add the few I know, but we'll have to revisit this whenever something else comes up. It shoudl be noted that LASYM has no effect on the structure of the file.

jlconlin (approved):

PR #89, opened 2020-10-21
Fixing g++ ranges compilation error

whaeck:
CI now works on release but I had to fix a ranges compilation error.

CI will by default trigger the release version, we'll need to fix that. And from now on, build/fetchcontent (aka develop for lack of a better name) is now a protected branch. Hurrah!

### DIFF
--- a/src/ENDFtk/section/6/ContinuumEnergyAngle/LegendreCoefficients/src/generateList.hpp
+++ b/src/ENDFtk/section/6/ContinuumEnergyAngle/LegendreCoefficients/src/generateList.hpp
@@ -1,5 +1,5 @@
 static std::vector< double >
-generateList( unsigned int na, 
+generateList( unsigned int na,
               std::vector< double >&& energies,
               std::vector< std::vector< double > >&& coefficients ) {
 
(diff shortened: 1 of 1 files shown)


src/ENDFtk/section/6/ContinuumEnergyAngle/LegendreCoefficients/src/generateList.hpp
```diff
@@ -27,5 +27,5 @@ generateList( unsigned int na,
   return ranges::view::zip_with(
              ranges::view::concat,
              energies | ranges::view::transform( ranges::view::single ),
-             coefficients ) | ranges::view::join;
+             coefficients ) | ranges::view::join | ranges::to_vector;
```
nathangibson14, line 30: Shouldn't this be automatic?  I guess this is a bug in the compiler and/or ranges?

nathangibson14 (approved):

src/ENDFtk/section/6/ContinuumEnergyAngle/LegendreCoefficients/src/generateList.hpp
whaeck: I don't know what's going on. Since the return type is a vector, it should be clear for the compiler. It's not the first time I've seen this.

whaeck:
And this will be how we'll do things from now on ;-)

PR #90, opened 2020-10-21
One Library to rule them all, and in python bind them

whaeck:
This is the pull request to integrate the python bindings of ENDFtk into the build/fetchcontent (aka develop branch for lack of a better name) branch.

The basis for these python bindings were @nathangibson14 prototype bindings (on the pybind branch) from which we learned a lot on how we should go about making these bindings. The main problems that were identified are the use of vectors whenever a range is returned and the requirement of a patch file to implement a better looking interface (especially with the parse< x, y > functions on the tree). The second one has been solved by using variant types and lambdas but the first one needed a more elaborate solution.

Whenever a range was encountered, the prototype bindings dumped everything into a vector (essentially copying the data whenever a property/function was called). As a solution, we have now implemented this using template type erasure with the ranges::view::any_view. Some preliminary ACEtk python bindings developed internally at LANL by a student and @david8dixon illustrated the use of this technique. The one implemented here is slightly different from that one. Whenever the range contains elements of a non-primitive type, the any_view will be to a reference_wrapper of that non-primitive type. When chaining python lookup of objects in a range of ranges, the reference_wrapper fixed the issues with the object going out of scope. Ranges of primitive types (in, double, etc.) are implemented as any_views of those primitive type since using a reference makes no sense in that case. All any_view wrapping utilities can be found in the views.hpp. The any_view can be converted into a python list either by using list( ... ) or the to_list() method on the bound any_view. In their current implementation, the any_view works for things like list and dict comprehension. Please note that not all any_view can be subscripted with [...] (for instance: values() or keys() ranges coming out of C++ maps are not random access).

Repetitive code is also implemented as templated functions, e.g. for wrapping recurring methods like NC, to_string, ... This applies to common TAB2, TAB1, section and file methods. These helper functions can be found in definitions.hpp, print.hpp and read.hpp.

The main module is called ENDFtk (sorry @jlconlin, no pie for you) and has been organised into submodules: one for every MF number (e.g. MF1, MF2, etc.) and the ENDF tree components that allow us to parse files and sections (obviously, that one is called tree). Whenever required (e.g. when a given MF has different types of MT sections, or for LAW components in MF6), additional submodules are introduced. Every ENDF section is called a Section and every file is called a File. The different types are distinguished using their respective submodule:
```python
import ENDFtk

tape = ENDFtk.tree.Tape.from_file( 'somefile' )
mf1mt451 = tape.material( 125 ).file( 1 ).section( 451 ).parse()
mf3mt102 = tape.material( 125 ).file( 3 ).section( 102 ).parse()
mf3 = tape.material( 125 ).file( 3 ).parse()

print( isinstance( mf1mt451, ENDFtk.MF1.MT451.Section ) )
print( isinstance( mf3mt102, ENDFtk.MF3.Section ) )
print( isinstance( mf3, ENDFtk.MF3.File ) )
```

Docstring documentation is provided for every component:
```python
import ENDFtk

help( ENDFtk )

tape = ENDFtk.tree.Tape.from_file( 'somefile' )
help( tape )
help( ENDFtk.MF1.MT451.Section )
```

The main source for this docstring documentation are the Doxygen comments already found in the C++ code. As a consequence, Doxygen documentation was added where none was initially provided (roughly 1/4 to 1/3 of the ENDFtk code base).

We took this opportunity to review the ENDFtk C++ interfaces and add/change/remove what we did not like and streamline the interfaces between components that offer a similar functionality (e.g TabulatedDistribution encountered in MF4 and multiple LAWs in MF6 use the same function/property names between them). The following components are the ones that have been changed and for which backwards compatibility WILL break:
- MF4, MF5 and MF6: renamed "subsections()" to "distributions()" (subsections was a leftover name for when the distribution variant was stored in a SubSection class, that was removed some time ago) and changed the name of quite a few classes to more logical names
- MF7: specific components have been tagged for change but the changes have NOT been implemented since MF7 is already in use for thermr and leapr.

These python bindings cover EVERYTHING that is currently in ENDFtk (with the exception of MF8 MT457) and this includes constructors for every section and component (the prototype bindings only allowed for accessing the data, not constructing it). The following is thus available: MF1 MT451, MF2, MF3, MF4, MF5, MF6, MF7, MF12 and MF13. From now on, new ENDF components WILL have python bindings added at the same time.

Python unit tests for every component that is bound have been provided (these are essentially translations of the C++ unit tests for the components) to show that all python bindings work as advertised. There are currently 89 test files (for a total of 167 test functions). For comparison: C++ ENDFtk has 126 unit test files.

Some building instructions:
```
git clone https://github.com/njoy/ENDFtk
git checkout feature/python-v2
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release ..
make ENDFtk.python -j4
```
This will only build the python bindings. To run the unit tests (assuming you're in the 'python' folder):
```
cp ../build/*.so . # you can also use a symbolic link
python -m unittest discover -v -p "Test*"
```

@jlconlin and @nathangibson14 have been tagged as reviewers for this PR. Don't be scared of the size, it will only hurt a bit ...

@staleyLANL I did not add you, but you may want to have a look to see how pybind11 works for making python bindings.

We may want to schedule a short meeting to provide a more digestible review process. It'll allow us to properly introduce the chosen interface style, the any_view solution and other problems that were encountered (and fixed) related to how pybind11 works (e.g. variant types needing default constructors and move semantics in C++ constructors).


whaeck:
BTW: CI has a problem with ranges again. Everything compiles using clang, g++-7 and g++-9 on my Mac and personal laptop.

staleyLANL:
Are you competing with me for writing long emails? :-D  Of course it was a PR comment, not directly an email. Seriously, this looks like good work, and thanks for the detailed description. This might benefit from a presentation, where you break it down into bite-sized pieces.

whaeck:
@staleyLANL You know I'm a competitive person ;-) Joke aside, there's a lot in here and it merits some initial explanation before you dive in. And yes, I agree that a presentation of specific pieces is of interest here.

PR #93, opened 2020-11-16
Feature/xy

whaeck:
Added common convenience interface to components that inherit or look and feel like TAB1 in MF3, MF12 and MF13.

CI fails on this due to a ranges issue with gcc 9.3. It does compile using clang.


src/ENDFtk/section/12/TotalMultiplicity.hpp
```diff
@@ -39,6 +39,9 @@ class TotalMultiplicity : protected TabulationRecord {
   using TabulationRecord::NP;
   using TabulationRecord::interpolants;
   using TabulationRecord::boundaries;
+  using TabulationRecord::x;
+  using TabulationRecord::y;
```
jlconlin, line 43: What's the difference between `y` and `Y`?

jlconlin (approved):

src/ENDFtk/section/12/TotalMultiplicity.hpp
whaeck: Y (as in capital Y) is the ENDF variable name for the multiplicities in MF12. y is just the convenience function from the TAB1, which in this case point to the same function.

PR #102, opened 2020-12-03
Feature/lthr3

whaeck:
The ENDF format for MF7/MT2 has changed: an additional option for the elastic scattering law has been added (LTHR=3 or mixed elastic scattering).

This adds the MixedElastic MF7/MT2 component, composed of a CoherentElastic and IncoherentElastic component. Python coding is added as well, along with both C++ and python unit tests.


src/ENDFtk/section/7/2/MixedElastic/src/NC.hpp
```diff
@@ -0,0 +1,7 @@
+/**
+ *  @brief Return the number of lines in this MF7/MT2 component
+ */
+long NC() const {
+
+  return this->coherent_.NC() + this->incoherent_.NC();
```
nathangibson14, line 6: I think this is right, but I thought I'd leave a comment to double check.

Does `NC` include the header lines?  If it did, this would double count the first two lines.  But I think `NC` is just the "meat" of the section.

src/ENDFtk/section/7/2/MixedElastic/src/print.hpp
```diff
@@ -0,0 +1,16 @@
...
+ *  @param[in] it            the current position in the output
+ *  @param[in] MAT           the MAT number
+ *  @param[in] MF            the MF number
+ *  @param[in] MT            the MT number
+ */
+template< typename OutputIterator >
+void print( OutputIterator& it, int MAT, int MF, int MT ) const {
+
+  this->coherent_.print( it, MAT, MF, MT );
+  this->incoherent_.print( it, MAT, MF, MT );
```
nathangibson14, line 15: Does this correctly set the `LTHR` flag and the rest of the header?

nathangibson14 (commented):
Looks good to me.  Just left a few likely unnecessary questions.  I'll make sure I can compile it and run the tests, then I'll approve.

src/ENDFtk/section/7/2/MixedElastic/src/NC.hpp
whaeck: If this had been a section, yes. In this case, this is merely a component so we just need to count the number of lines in the coherent and incoherent component that make up this mixed component.

nathangibson14: Aha, that's right.  I'm slow sometimes :smile:

src/ENDFtk/section/7/2/MixedElastic/src/print.hpp
whaeck: The LTHR flag is not set in these components, it's set on the HEAD record, in the section::Type<7,2>. The LTHR() functions do this work.

whaeck: BTW: this is the way we've handled all cases like this: static constexpr functions on the component to provide the format flag value associated to it, which are used in the components that contain them.

nathangibson14 (approved):

PR #108, opened 2020-12-16
Feature/kbk

whaeck:
Added the necessary components for MF2 MT151 KBK options: background R-matrix for LRF=7

New components have been added, ResonanceChannels and SpinGroup were modified/expanded (renamed the human readable function for KBK to relfect the change in meaning of KBK). The changes in existing components do not impact code in r2 since it was unused anyway.

Added/modified tests in C++ and python.


jlconlin:
What is KBK?

whaeck:
These are the background R-matrix options available in LRF=7. Previously, they could not be used since the ENDF format for them was ill-defined, most likely because the format was never properly finished/vetted (the NPL value of a LIST was used to specify the option: LBK=0,1,2,3 knowing that the data for options 2 and 3 consisted of 5 and 7 values). Jesse Brown from ORNL made an ENDF formats proposal to clean up LBK=2 at CSEWG in 2019 (it was conditionally approved, provided that all other options were cleaned up as well). Jesse, Doro and I then revised that proposal for CSEWG 2020 to enable all options.

These formats specify values R_cc to be added to the R-matrix diagonal in LRF=7 Reich-Moore using 4 options:
- nothing (LBK=0)
- tabulated complex function
- SAMMY or Frohner parametrisations (formulas for these have been in Appendix D of the ENDF manual for ages)

There currently are no evaluations using these options (obviously) although there will be at least ONE in ENDF/B-VIII.1 that will use the SAMMY parametrisation. Jesse Brown is working on that evaluation. The current implementation for resonance reconstruction is essentially equivalent to the first option.

src/ENDFtk/section/2/151/RMatrixLimited/BackgroundChannels.hpp
```diff
@@ -0,0 +1,60 @@
...
+   *  @brief Return the number of channels with background R-matrix values
+   */
+  auto KBK() const {
+
+    return ranges::count( this->backgroundRMatrices()
+                              | ranges::view::transform(
+                                    [] ( const auto& value )
+                                       { return bool( value ); } ),
+                          true );
+  }
```
nathangibson14, line 51: KBK is given in the ENDF file.  I can see calculating it being useful when writing a file.  But does it make sense not to use the one in the file on read?

nathangibson14: I think what I'm saying is -- why compute this with ranges when it could be stored (and incremented/decremented at creation)?

src/ENDFtk/section/2/151/RMatrixLimited/FrohnerBackgroundRMatrix.hpp
```diff
@@ -0,0 +1,83 @@
...
+ *  @brief Frohner parametrisation as a channel's background R-matrix
+ *
+ *  See ENDF102, section 2.2.1.6 for more information.
+ */
+class FrohnerBackgroundRMatrix : protected BaseBackgroundRMatrix< FrohnerBackgroundRMatrix > {
+
+  friend BaseBackgroundRMatrix< FrohnerBackgroundRMatrix >;
+
+  /* fields */
+  std::array< double, 5 > data_;
```
nathangibson14, line 12: Why an array internally instead of explicitly storing each parameter?

src/ENDFtk/section/2/151/RMatrixLimited/FrohnerBackgroundRMatrix/src/extract.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+  if ( record.NPL() != 3 ) {
+
+    Log::info( "Expected NPL = 3 for the Frohner parametrisation" );
+    Log::info( "Found NPL value: {}", record.NPL() );
+    throw std::exception();
+  }
+
+  auto list = record.list();
+  std::copy_n( ranges::begin( list ), 3, values.begin() + 2 );
```
nathangibson14, line 16: This is a really weird way of doing this.  Why not just `copy_n` the whole list after checking `NPL` first?  Doing two values, checking, then doing three more with a different procedure seems odd.

src/ENDFtk/section/2/151/RMatrixLimited/FrohnerBackgroundRMatrix/src/verifyLBK.hpp
```diff
@@ -0,0 +1,11 @@
+static void
+verifyLBK( int LBK ) {
+
+  if ( LBK != 3 ) {
```
nathangibson14, line 4: The ENDF format is stupid for putting LBK (and so many other quantities) in multiple places.

src/ENDFtk/section/2/151/RMatrixLimited/NoBackgroundRMatrix/src/ctor.hpp
```diff
@@ -0,0 +1,49 @@
...
+ */
+NoBackgroundRMatrix() = default;
+#endif
+
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] index    the channel index
+ */
+NoBackgroundRMatrix( int index ) : BaseBackgroundRMatrix( index ) {}
```
nathangibson14, line 14: If a user were to want to make a background matrix with KBK > 0, could we make it so they only specified the non-trivial channels and just automatically insert this behind the scenes?  Constructing a dummy class as a user seems unnecessary.

src/ENDFtk/section/2/151/RMatrixLimited/SammyBackgroundRMatrix.hpp
```diff
@@ -0,0 +1,93 @@
...
+ *  @brief SAMMY parametrisation as a channel's background R-matrix
+ *
+ *  See ENDF102, section 2.2.1.6 for more information.
+ */
+class SammyBackgroundRMatrix : protected BaseBackgroundRMatrix< SammyBackgroundRMatrix > {
+
+  friend BaseBackgroundRMatrix< SammyBackgroundRMatrix >;
+
+  /* fields */
+  std::array< double, 7 > data_;
```
nathangibson14, line 12: Same question I asked on Frohner -- why the array?

src/ENDFtk/section/2/151/RMatrixLimited/SammyBackgroundRMatrix/src/extract.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+  if ( record.NPL() != 5 ) {
+
+    Log::info( "Expected NPL = 5 for the SAMMY parametrisation" );
+    Log::info( "Found NPL value: {}", record.NPL() );
+    throw std::exception();
+  }
+
+  auto list = record.list();
+  std::copy_n( ranges::begin( list ), 5, values.begin() + 2 );
```
nathangibson14, line 16: Same comment as before -- this is weird to me.

src/ENDFtk/section/2/151/RMatrixLimited/BackgroundChannels.hpp
whaeck: KBK is stored on the ResonanceChannels, so we do want to know how many are defined in the BackgroundChannels so we can compare to the KBK value on the ResonanceChannels. This allows us to more easily test for consistency.

I use a similar construct in other places, such as the LAW value which is stored on the Multiplicity.

src/ENDFtk/section/2/151/RMatrixLimited/FrohnerBackgroundRMatrix.hpp
whaeck: It's an implementation detail ;-)

I could have done it that way, I'll have to check the code but I think I did that because the data comes out of a list so I can dump it in an array.

src/ENDFtk/section/2/151/RMatrixLimited/FrohnerBackgroundRMatrix/src/extract.hpp
whaeck: The data is actually dispersed on the list record: two values are stored in the C1 and C2 fields (and these are not part of the values in the list). Moving the assignment of the first two values in the array to after the NPL check does make sense since the assignment of the entire array would be in the same spot.

src/ENDFtk/section/2/151/RMatrixLimited/FrohnerBackgroundRMatrix/src/verifyLBK.hpp
whaeck: I have to work with what I got ;-)

src/ENDFtk/section/2/151/RMatrixLimited/NoBackgroundRMatrix/src/ctor.hpp
whaeck: The value of KBK actually gives the number of background channels the user defines so the user does not have to define a background channel for every resonance channel.

A user is allowed to write:
```
        chunk = BackgroundChannels(
                    nch = 3,
                    channels = [ SammyBackgroundRMatrix( 2, 1., 2., 3., 4., 5., 6., 7. ) ] )
```
to define a Sammy background channel for the channel with index 2, knowing that there are 3 channels in total. When no data is given for a specific channel, it defaults to NoBackgroundRMatrix.

He could also have written this to have the same effect:
```
        chunk = BackgroundChannels(
                    nch = 3,
                    channels = [ NoBackgroundRMatrix( 0 ),
                                 NoBackgroundRMatrix( 1 ),
                                 SammyBackgroundRMatrix( 2, 1., 2., 3., 4., 5., 6., 7. ) ] )
```

nathangibson14 (approved):
Tests pass successfully on my machine.

Issue #101, opened 2020-12-03
Implement python helper function like from_file on the C++ side as well


whaeck:
See @nathangibson14's comment in PR #92

Issue #100, opened 2020-11-24
Add a convenience function to get the first material in a TAPE


whaeck:
For example:
```
material = tape.mat()
material = tape.front()
```

As a side note: maybe add front(), back() functions to the view interfaces?

PR #109, opened 2020-12-16
Small extensions

whaeck:
Addressing issues #100 and #101

### DIFF
--- a/python/src/tree/Tape.python.cpp
+++ b/python/src/tree/Tape.python.cpp
@@ -5,7 +5,7 @@
 // local includes
 #include "ENDFtk/tree/Tape.hpp"
 #include "ENDFtk/tree/makeTape.hpp"
-#include "header-utilities/slurpFileToMemory.hpp"
+#include "ENDFtk/tree/fromFile.hpp"
 #include "views.hpp"
 
 // namespace aliases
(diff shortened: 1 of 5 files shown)


python/src/tree/Tape.python.cpp
```diff
@@ -5,7 +5,7 @@
 // local includes
 #include "ENDFtk/tree/Tape.hpp"
 #include "ENDFtk/tree/makeTape.hpp"
-#include "header-utilities/slurpFileToMemory.hpp"
+#include "ENDFtk/tree/fromFile.hpp"
```
nathangibson14, line 8: Aren't you going to miss "slurp"?

nathangibson14 (changes requested):
Shouldn't we be testing the `.front`?

Other than that, looks great.

python/src/tree/Tape.python.cpp
whaeck: Not in this lifetime ;-)

python/test/Test_ENDFtk_Tree_Tape.py
```diff
@@ -221,6 +221,9 @@ def verify_material( self, material ) :
             material = materials.to_list()[0]
             verify_material( self, material )
 
+            material = materials.front()
+            verify_material( self, material )
```
nathangibson14, line 225: Do you think testing in Python but not C++ is sufficient?  I'm on the fence.

whaeck: This is only available in python, not on the C++ side.

nathangibson14: That's a good point.  `.front` works out of the box in C++. 🤦

nathangibson14 (approved):

PR #110, opened 2020-12-17
Fix/views

whaeck:
A few updates:
- all views are now put into the ENDFtk.sequence submodule
- changed the wrapping function order to ensure the python documentation uses the python type names instead of the original C++ type names
- fixed some typos in the documentation

This branch merges all outstanding pull requests so it must be merged last.


python/src/record/DirectoryRecord.python.cpp
```diff
@@ -85,6 +86,7 @@ void wrapDirectoryRecord( python::module& module ) {
                 ( self.MT() == right.MT() ) and
                 ( self.NC() == right.NC() ) and
                 ( self.MOD() == right.MOD() ); },
+    python::arg( "right" ),
```
nathangibson14, line 89: I'm not convinced there's a use for this one, but might as well leave it for completeness!

python/src/section/1/455.python.cpp
```diff
@@ -15,21 +15,19 @@ namespace mf1 {
...
+void wrapSection_1_455( python::module& module, python::module& viewmodule ) {
 
   // type aliases
   using Section = njoy::ENDFtk::section::Type< 1, 455 >;
-  using DecayConstantData = njoy::ENDFtk::section::Type< 1, 455 >::DecayConstantData;
-  using Multiplicity = njoy::ENDFtk::section::Type< 1, 455 >::Multiplicity;
-
-  // wrap views created by this section
+  using DecayConstantData = Section::DecayConstantData;
+  using Multiplicity = Section::Multiplicity;
```
nathangibson14, line 30: I think I previously made a comment on these.  I really like the more compact notation this gives you!

python/src/section/1/TabulatedMultiplicity.python.cpp
```diff
@@ -10,7 +10,7 @@
 // namespace aliases
 namespace python = pybind11;
 
-void wrap_1_TabulatedMultiplicity( python::module& module ) {
+void wrap_1_TabulatedMultiplicity( python::module& module, python::module& ) {
```
nathangibson14, line 13: Are you putting in the dummy argument on these just to keep the interface the same on all of them?  That probably does make things easier...

python/src/section/2/151.python.cpp
```diff
@@ -11,55 +11,84 @@
...
+void wrapUnresolvedEnergyDependentFissionWidthsJValue( python::module&, python::module& );
+void wrapUnresolvedEnergyDependentJValue( python::module&, python::module& );
+void wrapUnresolvedEnergyIndependentLValue( python::module&, python::module& );
+void wrapUnresolvedEnergyDependentFissionWidthsLValue( python::module&, python::module& );
+void wrapUnresolvedEnergyDependentLValue( python::module&, python::module& );
+void wrapUnresolvedEnergyIndependent( python::module&, python::module& );
+void wrapUnresolvedEnergyDependentFissionWidths( python::module&, python::module& );
+void wrapUnresolvedEnergyDependent( python::module&, python::module& );
+void wrapResonanceRange( python::module&, python::module& );
+void wrapIsotope( python::module&, python::module& );
```
nathangibson14, line 40: Getting these in the right order must have been a fun exercise in topological sorting...

nathangibson14 (approved):
This looks great, and the result of a lot of tedious work -- thanks for doing it!

The one thing that's nearly impossible for a PR of this sort is to adjudicate the completeness of the change.  So let's make sure we keep an eye out for this as we beta-test our upcoming release candidate.

python/src/section/1/455.python.cpp
whaeck: Yes, you did ;-)

python/src/section/1/TabulatedMultiplicity.python.cpp
whaeck: Yes, exactly. That way, I don;t have to think about it, or change it later if I do need a view in that function.

whaeck:
I've actually gone over the help(  ) of every module and class in ENDFtk to verify. I think I caught everything.

I'll branch of the release branch once I have this in.

PR #111, opened 2020-12-22
Feature/tree

whaeck:
A few simple changes:
- enabled the copy constructor on the ENDFtk tree components on the python side
- added some simple python unit testing on this new feature

The copy constructor on the tree.Material. tree.File and tree.Section is compiler generated but Austin did add tests for the copy and move constructor on the C++ side. As such, I am convinced the feature works as advertised. Tests in the python interpreter did not reveal issues.

### DIFF
--- a/python/src/tree/File.python.cpp
+++ b/python/src/tree/File.python.cpp
@@ -50,8 +50,17 @@ void wrapTreeFile( python::module& module ) {
   );
 
   // wrap the tree component
-  // no __init__ since we do not want to create this object in python
+  // only copy is allowed since we do not want to create this object in python
   tree
+  .def(
+
+    python::init< const File& >(),
+    python::arg( "file" ),
+    "Initialise the file with another file\n\n"
+    "Arguments:\n"
+    "    self    the file\n"
+    "    file    the file to be copied"
+  )
   .def_property_readonly(
 
     "MAT",
--- a/python/src/tree/Material.python.cpp
+++ b/python/src/tree/Material.python.cpp
@@ -35,8 +35,17 @@ void wrapTreeMaterial( python::module& module ) {
   );
 
   // wrap the tree component
-  // no __init__ since we do not want to create this object in python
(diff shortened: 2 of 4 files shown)


python/src/tree/File.python.cpp
```diff
@@ -50,8 +50,17 @@ void wrapTreeFile( python::module& module ) {
   );
 
   // wrap the tree component
-  // no __init__ since we do not want to create this object in python
+  // only copy is allowed since we do not want to create this object in python
   tree
+  .def(
+
+    python::init< const File& >(),
```
nathangibson14, line 57: It's pretty neat that this is all it takes to link to the C++ constructors.

nathangibson14 (commented):
Nice and simple change!  Your testing does not quite cover the use case I had -- and that use case is hard to capture in a test.  I'll try it in my scripts, with the tree (and its buffer) going out of scope.  I'll approve if/when that works as expected.

python/src/tree/File.python.cpp
whaeck: Well, that (linking __init__ to a constructor signature) and lambdas if the first is impossible make this easy. pybind11 is quite a nice library.

whaeck:
Austin's test actually covers it:
```
        auto beg = begin;
        auto lin = lineNumber;
        auto original =
          std::make_unique
          < tree::Material< std::string::iterator > >
          ( head, start, beg, end, lin );

        THEN( "the copy ctor will function correctly "){
          auto copy = tree::Material< std::string::iterator >{ *original };
          original.reset();
          REQUIRE( 3 == copy.MF( 3 ).MF() );
        }
```

As you can see, he creates a unique pointer to a material, copies it and then resets the pointer (essentially erasing the original material). The copy on the other hand still contains the information.

nathangibson14 (approved):
Looks good to me.  Glad that there is a test for it.  Can confirm it works as desired for my purposes.

PR #117, opened 2021-02-01
Updated readme with some troubleshooting information

whaeck:


nathangibson14 (approved):
I don't really agree with your capitali**z**ation choices with things like python and linux, but otherwise this looks good.

whaeck:
We'll fix it later ;-)

PR #119, opened 2021-02-19
GENDF -- 8 months later

nathangibson14:
Here is the start of GENDF capabilities reimagined.  This represents a good start, but there's quite a bit more to do.  `CrossSection` still needs constructors from parameters.  Then, there will be sister classes to `CrossSection` called `Ratio` and `Matrix`, plus possibly `DelayedFission`.  And MFD1 will need special treatment.  Printing is yet to come.

Syntax tree/parse is further down the road, but it'll come!

But I'd rather get some feedback from @whaeck before I get too far on this one.


whaeck:
@nathangibson14 You're aware that the tests seem to fail?

nathangibson14:
I saw that Actions failed, but I have no idea why.  The tests pass for me locally.  I've made enough changes, though, that it's not worth figuring out what happened here for now.  I'll keep an eye on it down the road.

PR #120, opened 2021-02-20
Feature/mf10

whaeck:
New stuff: MF9 and MF10 data ;-)

PR order:
#123
#120 <-- this one
#124 
#126
#127

PR #124, opened 2021-03-22
Feature/mf14

whaeck:
Full MF14 support.

PR order:
#123
#120
#124 <-- this one 
#126
#127


src/ENDFtk/section/14/src/readPhotons.hpp
```diff
@@ -0,0 +1,54 @@
...
+             long NI,
+             long NK,
+             long LTT) {
+
+  std::vector< PhotonDistribution > sequence;
+  if ( LI != 1 ) {
+
+    sequence.reserve( NK );
+
+    NK -= NI;
```
nathangibson14, line 19: Especially where you put this line requires a comment.  I was confused for a second, but I agree this is correct.

src/ENDFtk/section/14.hpp
```diff
@@ -0,0 +1,142 @@
...
+     *  @brief Return the number of photons with an isotropic angular
+     *         distribution
+     */
+    int numberIsotropicPhotons() const { return this->NI(); }
+
+    /**
+     *  @brief Return the number of photons (discrete and continuum) with
+     *         angular distributions
+     */
+    int NK() const { return this->photons_.size(); }
```
nathangibson14, line 89: For `LI=0`, `NK` need not be zero (and in fact technically shouldn't be), but `photons` will be an empty list.  This is a failing of the ENDF format, as `NK` is redundant information in that case.  But alas...

src/ENDFtk/section/14.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     *  @brief Return the isotropic angular distribution flag
+     */
+    bool LI() const { return this->photons_.size() == 0; }
+
+    /**
+     *  @brief Return the isotropic angular distribution flag
+     */
+    bool isotropicAngularDistributions() const { return this->LI(); }
```
nathangibson14, line 105: I don't like this variable name.  It sounds like it'll return a list of the isotropic distributions, when really it's just a flag.  ENDF doesn't give a name we need to follow.  Maybe `isotropicFlag` or `distributionsIsotropic`?  Pedantic, but might reduce confusion.

src/ENDFtk/section/14.hpp
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     *  @brief Return the distribution law
+     */
+    int LTT() const {
+
+      return this->LI() ? 0
+                        : std::visit( [] ( const auto& photon )
+                                         { return photon.LTT(); },
+                                      this->photons_.back() );
+    }
```
nathangibson14, line 116: This works, but ehhh.  This seems like something I'd just store and verify rather than computing.  But if it works, there's no real reason to change it.

src/ENDFtk/section/14.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     *  @brief Return the distribution law
+     */
+    int LAW() const { return this->LTT(); }
+
+    /**
+     *  @brief Return the secondary photons with their angular distribution
+     */
+    const auto& photonAngularDistributions() const { return this->photons_; }
```
nathangibson14, line 126: This name is so similar to `isotropicAngularDistributions` with a totally different meaning, +1 for my previous comment.

src/ENDFtk/section/14/TabulatedDistribution.hpp
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   *  @brief Return the incident energy for which the angular distribution
+   *         is given
+   */
+  double incidentEnergy() const { return this->E(); }
+
+  /**
+   *  @brief Return the cosine values
+   */
+  auto MU() const { return TabulationRecord::x(); }
```
nathangibson14, line 37: You're doing the right thing, but `MU` as a variable name seems so wrong.  Silly ENDF-speak.

src/ENDFtk/section/14/LegendreDistributions.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   *  @brief Return the isotropic angular distribution flag
+   */
+  bool isotropicAngularDistributions() const { return this->LI(); }
+
+  /**
+   *  @brief Return the distribution law
+   */
+  static constexpr int LTT() { return 1; }
```
nathangibson14, line 60: I don't like defining `LI` and `LTT` for a single distribution, it's defined on the section level.  These quantities aren't defined at this level in ENDF, there are just 0s where you could imagine them being.

src/ENDFtk/section/14/TabulatedDistributions.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   *  @brief Return the isotropic angular distribution flag
+   */
+  bool isotropicAngularDistributions() const { return this->LI(); }
+
+  /**
+   *  @brief Return the distribution law
+   */
+  static constexpr int LTT() { return 2; }
```
nathangibson14, line 60: Same comment as with LegendreDistributions.  I don't think this belongs here.

src/ENDFtk/section/14/TabulatedDistributions.hpp
```diff
@@ -0,0 +1,76 @@
...
+ *  @brief Angular distributions as a function of incident energy using
+ *         tabulated functions (LTT=2) for a given discrete photon
+ *
+ *  The LegendreDistributions class is used to represent the case in which
+ *  the angular distributions in an MF4 section are given as tabulated
+ *  functions.
+ *
+ *  See ENDF102, section 14.2.3 for more information.
+ */
+class TabulatedDistributions :
```
nathangibson14, line 12: I don't think this should be pluralized.  It's a TAB2 that you are representing, which I'd argue is just one `TabulatedDistribution`.  Likewise for Legendre.

src/ENDFtk/section/4/Isotropic.hpp
```diff
@@ -15,6 +15,16 @@ class Isotropic {
...
 
+  /**
+   *  @brief Return the isotropic angular distribution flag
+   */
+  static constexpr bool LI() { return true; }
+
+  /**
+   *  @brief Return the isotropic angular distribution flag
+   */
+  bool isotropicAngularDistributions() const { return this->LI(); }
```
nathangibson14, line 26: Don't like this variable name here either :smile:

python/src/section/14/TabulatedDistribution.python.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+    python::init< double, std::vector< long >&&, std::vector< long >&&,
+                  std::vector< double >&&, std::vector< double >&& >(),
+    python::arg( "energy" ), python::arg( "boundaries" ),
+    python::arg( "interpolants" ), python::arg( "cosines" ),
+    python::arg( "probabilities" ),
+    "Initialise the component\n\n"
+    "Arguments:\n"
+    "    self             the component\n"
+    "    incident         the incident energy value\n"
```
nathangibson14, line 43: Inconsistent variable name and argument.

nathangibson14 (changes requested):
Left a few comments that should be addressed.

Bigger picture, we may need to think of an easier interface for creating an `LI != 0` section.  Making a `IsotropicDiscreteDistribution` for each isotropic photon seems like overkill.  Maybe something where just a list of isotropic energies is given?

src/ENDFtk/section/14.hpp
whaeck: Well, technically it's a function name, not a variable :-)

What about isotropicDistributionFlag, since angular is understood by the isotropic in there.

whaeck: Maybe better isotropicDistributions() so that we can write:
```
if not data.isotropic_distributions :
    # do stuff
else :
    probability = 0.5
```

nathangibson14: I can deal with `isotropicDistributions`, as it's at least significantly different from `photonAngularDistributions`.  But I think I like the option with `Flag` better.

src/ENDFtk/section/14/LegendreDistributions.hpp
whaeck: This is similar to how some of the other sections work. The resonance formalism LRF value for instance is normally stored in the resonance region but we only store what is relevant and have each formalism component returns its LRF value when we need to using a static constexpr function (or a normal function when there's multiple possibilities). The MF6 sections use similar constructs with LAW/LANG.

python/src/section/14/TabulatedDistribution.python.cpp
whaeck: Good catch.

src/ENDFtk/section/14/TabulatedDistributions.hpp
whaeck: There already is a TabulatedDistribution - which is a single tabulate distribution for a specific energy (which is also used as the underlying distribution type in the TAB2).

As with some of the other remarks: similar names are used in MF6 as well. If we change this here, we'll have to change it everywhere. Keeping a consistently name interface is hard if you don't develop all of it at the same time.

nathangibson14 (approved):
Looks good now!

PR #126, opened 2021-04-06
Feature/mf15

whaeck:
Full MF15 capability

PR order:
#123
#120 
#124 
#126  <-- this one
#127


src/ENDFtk/section/15.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+  /* constructor */
+  #include "ENDFtk/section/15/src/ctor.hpp"
+
+  /* methods */
+
+  /**
+   *  @brief Return the number NK of subsections with partial distributions
+   */
+  int NK() const { return this->partials_.size(); }
```
nathangibson14, line 55: This is called `NC` in the ENDF manual.

nathangibson14: ...which is super confusing, given that `NC` is usually the number of lines.  I don't know what the right thing to do is.

src/ENDFtk/section/15/PartialDistribution/src/verifyLF.hpp
```diff
@@ -0,0 +1,13 @@
+static void
+verifyLF( int partialLF, int spectrumLF ) {
+
+  if ( partialLF != spectrumLF ) {
```
nathangibson14, line 4: `LF` is only given on the probabilities according to the ENDF manual.  Where is `spectrumLF` coming from? I don't even see `LF` given on the spectrum information in MF5.

nathangibson14: Nevermind, I see it now.  You are confirming the probability's `LF` flag matches the `TabulatedSpectrum` type.  I vote in favor of a comment explaining this isn't two separate values.

python/src/section/15/Probability.python.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    python::init< long, std::vector< long >&&, std::vector< long >&&,
+                  std::vector< double >&&, std::vector< double >&&, double >(),
+    python::arg( "lf" ), python::arg( "boundaries" ),
+    python::arg( "interpolants" ), python::arg( "energies" ),
+    python::arg( "probabilities" ), python::arg( "u" ) = 0.0,
+    "Initialise the component\n\n"
+    "Arguments:\n"
+    "    self             the component\n"
+    "    lf               the type of the underlying distribution\n"
```
nathangibson14, line 43: Required arguments that only can take one value seems like a poor interface decision.  I get that this matches MF5 and the ENDF manual leaves open the possibility of eventually adding more.  But what are your thoughts on axing `lf` as in input?

nathangibson14 (commented):
I looked at at least most of the MF15 changes, but it's hard to know if there were any changes in other files that weren't in the MF14 PR.  Once that one is merged, I'll re-check.

This looks good overall.  Left a few comments, but nothing big.

nathangibson14 (approved):

PR #127, opened 2021-04-08
Removing range-v3-adapter and update range-v3 to the latest 0.11.0 version

whaeck:
This updates range-v3 and seems to resolve the g++-8 and g++-9 compilation problems.

PR order:
#123
#120 
#124 
#126
#127  <-- this one


nathangibson14 (approved):
I left a few questions, but nothing looks worrisome.  This was a huge effort and was long overdue, so thanks for taking one for the team by doing it!

I don't love the transition away from `auto`, as I think it makes the code worse to look at.  But I guess this gives us confidence that `auto` isn't an issue in these cases, and there's definitely no reason to revert.  So I say let's stick with it.

PR #136, opened 2021-07-01
Python/records

whaeck:
Adding python bindings for CONT, LIST, TAB1 and TAB2.

### DIFF
--- a/python/src/record/ListRecord.python.cpp
+++ b/python/src/record/ListRecord.python.cpp
@@ -3,35 +3,37 @@
 #include <pybind11/stl.h>
 
 // local includes
-#include "ENDFtk/ControlRecord.hpp"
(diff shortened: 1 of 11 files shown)


python/src/record/ListRecord.python.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    "N2",
+    [] ( const Record& self ) { return self.N2(); },
+    "The integer in column 6"
+  )
+  .def_property_readonly(
+
+    "NPL",
+    &Record::NPL,
+    "The integer in column 6"
```
nathangibson14, line 90: Incorrect documentation.

nathangibson14 (changes requested):
One minor documentation issue, otherwise good.  Thanks for adding this for me!

nathangibson14 (approved):

PR #137, opened 2021-07-08
Rework/tree

whaeck:
The reworked ENDF tree components: each Section now owns its buffer to allow for easy insertion and deletion of sections, files, materials, etc.

The tree components are no longer templated, only some of the constructors are.

The tree File and Material now have functions to remove, insert and replace tree Section and File instances. The tree Section also has an NC function to return the number of lines (required to reconstitute a new index for a material).

Tests for the Material, File and Section were refactored and tests were simplified (due to the introduction of the new Section constructor).

There is a slight reduction in parsing performance when parsing a material in its entirety. The new tree implementation takes about 10% longer. For example: parsing 556 incident neutron files now takes about 13.3 seconds while it used to take 12 seconds. There's some room for optimisation there.

There's a few more things that could have been implemented, but there's more than enough already to merit this pull request. The outstanding work is as follows:
- cleaning up the ENDF sequence numbers at the end of the line
- reindexing the material and updating the MF1 MT451 index
- optimising parsing the material and file

UPDATE July 15, 2021:
- reindexing is now enabled
- the sequence numbers can now be cleaned
- the tree components now allow for insertion, replacement and removal of specific pieces


src/ENDFtk/tree/updateDirectory.hpp
```diff
@@ -0,0 +1,86 @@
...
+namespace ENDFtk {
+namespace tree {
+
+  /**
+   *  @brief Update the MF1 MT451 directory for the given material
+   *
+   *  An exception will be thrown if the MF1 MT451 section is not present, or if
+   *  there was an issue parsing it.
+   */
+  inline void updateDirectory( Material& material ) {
```
nathangibson14, line 20: Why are these free functions instead of member functions in tree/Material and tree/Tape?

nathangibson14 (dismissed):
Reviewed the code, all looks good.  Verified everything works as expected for ugly GENDF files (within reason), double MENDs, etc.  Insert/remove/replace all work in my testing beautifully.

I'll rebase MF33 once this is into develop.

src/ENDFtk/section/14/test/14.test.cpp
```diff
@@ -78,6 +79,28 @@ SCENARIO( "section::Type< 14 >" ) {
...
       } // THEN
     } //WHEN
+
+    WHEN( "there is a tree::Section" ) {
+
+      tree::Section section( 9228, 14, 2, std::string( sectionString ) );
+
+      section::Type< 14 > chunk = section.parse< 14 >();
+
+      THEN( "a section::Type< 4 > can be constructed and members can be "
```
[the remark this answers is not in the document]
whaeck: typos ...

src/ENDFtk/tree/updateDirectory.hpp
whaeck: Because that way I don't have to include anything for the parsed sections (even the parsing is templated).

nathangibson14 (approved):

PR #138, opened 2021-07-08
Feature/mf23

whaeck:
MF23 is now added. MF23 is quite simple and almost identical to MF3 (there's no LR flag and C1 and C2 on the TAB1 have different meanings).

This builds on #137 so that one needs to be merged first.


nathangibson14 (approved):
Nice and easy file for sure.  Looks good to me, tests pass, good to go!

PR #139, opened 2021-07-15
Feature/mf27

whaeck:
MF27 is now added. Like MF23 it is quite simple and almost identical to MF3.

This builds on #138 so that one needs to be merged first.


src/ENDFtk/section/27.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     *  @brief Return the energy values
+     */
+    auto energies() const { return this->table.x(); }
+
+    /**
+     *  @brief Return the form factor or scattering function values
+     */
+    auto values() const { return this->table.y(); }
```
nathangibson14, line 72: Should you include the ENDF-speak of `H` for this?

nathangibson14 (changes requested):
Because MF27 data can mean different things, I'm a little iffy on the interface.  I think I'm okay with `energies` being something that can be returned, but I'd definitely change the constructor to be the more general `x`.

That all assumes that x isn't necessarily an energy.  The equations in the ENDF manual don't really make sense, as x appears without definition, and I don't know the physics enough to intuit it.  Do you know any more?

src/ENDFtk/section/27.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     *  @brief Return the interpolation boundaries
+     */
+    auto boundaries() const { return this->table.boundaries(); }
+
+    /**
+     *  @brief Return the energy values
+     */
+    auto energies() const { return this->table.x(); }
```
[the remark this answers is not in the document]
whaeck: I've replaced this with X and H, no human readable version.

src/ENDFtk.hpp
```diff
@@ -19,8 +19,14 @@
...
 #include "ENDFtk/section/7.hpp"
 #include "ENDFtk/section/8.hpp"
+#include "ENDFtk/section/9.hpp"
+#include "ENDFtk/section/10.hpp"
 #include "ENDFtk/section/12.hpp"
 #include "ENDFtk/section/13.hpp"
+#include "ENDFtk/section/14.hpp"
+#include "ENDFtk/section/15.hpp"
+#include "ENDFtk/section/23.hpp"
+#include "ENDFtk/section/27.hpp"
```
nathangibson14, line 29: I don't know if this is my fault or yours, but this is missing 33.hpp

src/ENDFtk.hpp
```diff
@@ -31,8 +37,14 @@
...
 #include "ENDFtk/file/7.hpp"
 #include "ENDFtk/file/8.hpp"
+#include "ENDFtk/file/9.hpp"
+#include "ENDFtk/file/10.hpp"
 #include "ENDFtk/file/12.hpp"
 #include "ENDFtk/file/13.hpp"
+#include "ENDFtk/file/14.hpp"
+#include "ENDFtk/file/15.hpp"
+#include "ENDFtk/file/23.hpp"
+#include "ENDFtk/file/27.hpp"
```
nathangibson14, line 47: Also missing 33.

nathangibson14 (changes requested):
One last fix needed, I'll just do it myself then approve.

nathangibson14 (approved):

PR #140, opened 2021-07-16
Feature/mf28

whaeck:
And MF28: atomic relaxation data

Tree, MF23 and MF27 pull requests should be done first


src/ENDFtk/section/28/SubshellData/Transition.hpp
```diff
@@ -0,0 +1,57 @@
...
+ *  @class
+ *  @brief Convenience interface for a transition
+ *
+ *  See ENDF102, section 28.2 for more information.
+ */
+template < typename Range >
+class Transition {
+
+  /* fields */
+  Range chunk;
```
nathangibson14, line 11: I laud your commitment to putting `chunk` in the actual code and not just tests.

src/ENDFtk/section/28/SubshellData/src/generateList.hpp
```diff
@@ -0,0 +1,34 @@
...
+               "are not consistent" );
+    Log::info( "NTR value: {}", ntr );
+    Log::info( "secondary.size(): {}", secondary.size() );
+    Log::info( "tertiary.size(): {}", tertiary.size() );
+    Log::info( "energies.size(): {}", energies.size() );
+    Log::info( "probabilities.size(): {}", probabilities.size() );
+    throw std::exception();
+  }
+
+  std::vector< double > list = { ebi, static_cast< double >( eln ), 0.,
```
nathangibson14, line 22: Do you really need to static cast an integer to a double?  Is it just because of the unsigned?  I thought int/double casting didn't need this.

python/src/section/28/SubshellData.python.cpp
```diff
@@ -0,0 +1,166 @@
...
+  component
+  .def(
+
+    python::init< unsigned int, double, unsigned int,
+                  std::vector< unsigned int >&&,
+                  std::vector< unsigned int >&&,
+                  std::vector< double >&&,
+                  std::vector< double >&& >(),
+    python::arg( "subshell" ), python::arg( "energy" ), python::arg( "eln" ),
+    python::arg( "secondary" ) = 0., python::arg( "tertiary" ) = 0,
```
nathangibson14, line 44: Why the `= 0` here?  These aren't optional on the C++ side.

nathangibson14 (commented):
Looks good.  Left a few questions.

src/ENDFtk/section/28/SubshellData/Transition.hpp
whaeck: I aim to please ;-)

src/ENDFtk/section/28/SubshellData/src/generateList.hpp
whaeck: Initialiser lists (the { ... } way of constructing objects) do not allow for implicit conversion so a cast is required. You actually get a compiler error.

python/src/section/28/SubshellData.python.cpp
whaeck: Yes, that doesn't look right. Probably a copy/paste error on my part.

nathangibson14 (approved):

PR #142, opened 2021-08-06
Feature/mf34

nathangibson14:
Full support for MF34.  MF34 is fairly similar to MF33, adding an extra layer `LegendreBlock` underneath `ReactionBlock` before the data.  The data itself can only be NI-Type, NC-Type is not supported.


python/test/Test_ENDFtk_MF33_ReactionBlock.py
```diff
@@ -74,6 +74,10 @@ def verify_chunk( self, chunk ) :
 
             self.assertEqual( 9, chunk.NC )
 
+            # verify string
+            self.assertEqual( self.chunk,
+                              chunk.to_string( 9437, 33, 2 ) )
```
whaeck, line 79: I had not noticed this was missing

cmake/unit_testing.cmake
```diff
@@ -117,6 +117,9 @@ add_subdirectory( src/ENDFtk/section/33/ReactionBlock/test )
 add_subdirectory( src/ENDFtk/section/33/RectangularMatrix/test )
 add_subdirectory( src/ENDFtk/section/33/SquareMatrix/test )
 add_subdirectory( src/ENDFtk/section/33/test )
+add_subdirectory( src/ENDFtk/section/34/LegendreBlock/test )
+add_subdirectory( src/ENDFtk/section/34/ReactionBlock/test )
+add_subdirectory( src/ENDFtk/section/34/test )
```
whaeck, line 122: When I reused components from another MF, I still copied over the test for that component in the new MF. I know it is (somewhat) redundant but it adds completeness to the testing. Same on the python side.

whaeck (approved):
Looks good. The comments are not show stoppers. The only one important enough to point out is the one about LCT, for consistency with other MF it would be better to use referenceFrame.

Since there's nothing really blocking, I've approved this so that you can merge once you've made changes you wish to make.

python/src/Material.python.cpp
```diff
@@ -48,7 +52,7 @@ void wrapMaterial( python::module& module, python::module& ) {
   // predefined lambda
   auto getFile = [] ( const Material& self, int mf )
   -> std::variant< MF1, MF2, MF3, MF4, MF5, MF6, MF7, MF8, MF9, MF10,
-                   MF12, MF13, MF14, MF15, MF33 > {
+                   MF12, MF13, MF14, MF15, MF23, MF27, MF28, MF33, MF34 > {
```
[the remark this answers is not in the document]
whaeck: It's actually a different variant with ref wrappers so disregard my comment. I wouldn't be against moving the definition of the ref wrapper variant in variant.hpp as well.

src/ENDFtk/section/34/ReactionBlock.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   *  @brief Return the material number for the second cross section (must be 0)
+   */
+  static constexpr int MAT1() { return 0; }
+
+  /**
+   *  @brief Return the material number for the second cross section (must be 0)
+   */
+  int secondMaterialNumber() const { return this->MAT1(); }
```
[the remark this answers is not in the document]
nathangibson14: It's in the ENDF manual as a reserved space, allowing for this to be changed in the future, presumably.  I don't think having a hook to it, even if returning a trivial 0, is a bad thing.

src/ENDFtk/section/34/src/print.hpp
```diff
@@ -0,0 +1,18 @@
...
+ *  @param[in] MAT           the MAT number
+ *  @param[in] MF            the MF number
+ */
+template< typename OutputIterator >
+void print( OutputIterator& it, int MAT, int MF ) const {
+
+  int MT = this->MT();
+  ControlRecord( this->ZA(), this->AWR(), 0,this->LTT(),
+                 0, this->NMT1() ).print( it, MAT, MF, MT );
+  for ( auto& entry : this->reactions_ ) { entry.print( it, MAT, MF, MT ); }
```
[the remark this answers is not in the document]
nathangibson14: I definitely blame you.  I'll fix a few of these, but I'm sure I won't get them all.  And it really doesn't matter anyway :)

cmake/unit_testing.cmake
nathangibson14: I'm not convinced this is worth it, but I can acquiesce.

src/ENDFtk/section/34/src/print.hpp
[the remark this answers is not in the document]
whaeck: I'll have to check then and fix it everywhere else. If you're going to correct it, only do this for MF3x. I'll fix it elsewhere.

It does not make a difference, but it's a const correctness thing that I try to adhere to.

cmake/unit_testing.cmake
whaeck: Technically speaking they are different types, since they are included inside a different class so technically they require testing but its a bit of a grey area. It's just for completeness. I do wonder what the line coverage would report in this case.

nathangibson14: I'd argue they are well-enough implicitly tested through their instantiation in the higher-level components.  But I added them, so I'm not sure why I'm arguing.

nathangibson14:
Should be all set after your review.  Tests pass locally.  You need to hit Approve again, I can't re-use your formal approval.

whaeck (approved):

PR #143, opened 2021-08-18
Feature/mf26

whaeck:
And ... I'm done! Full MF26 capability ;-)


src/ENDFtk/section/26/Multiplicity/src/ctor.hpp
```diff
@@ -0,0 +1,49 @@
...
+Multiplicity( double zap, double awi, long law,
+              std::vector< long >&& boundaries,
+              std::vector< long >&& interpolants,
+              std::vector< double >&& energies,
+              std::vector< double >&& multiplicities )
+  try : TabulationRecord( zap, awi, 0, law,
+                          std::move( boundaries ),
+                          std::move( interpolants ),
+                          std::move( energies ),
+                          std::move( multiplicities ) ) {}
```
nathangibson14, line 21: The ENDF manual says that the multiplicities must always be 1, not just "normally one".  Thus, it doesn't really make sense to require it as an input.  I think the only value of this constructor would be to make a user specify the energy range and the law.

That said, maybe consistency with MF6 is desired?  So maybe this is okay?

src/ENDFtk/section/26/Multiplicity/test/Multiplicity.test.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+  CHECK( 3 == chunk.NC() );
+}
+
+std::string invalidChunk() {
+
+  return
+  " 1.100000+1 5.438673-4          0          2          2          2 10026525     \n"
+  "          2          2                                             10026525     \n"
+  " 1.000000+1 1.000000+0 1.00000+11 1.000000+0                       10026525     \n";
```
nathangibson14, line 162: If we're going to leave multiplicity as a user input, we should have a `checkMultiplicity` function that makes sure it's all 1s.

src/ENDFtk/section/26/src/ctor.hpp
```diff
@@ -0,0 +1,42 @@
+/**
```
nathangibson14, line 1: Missing the Python default constructor.

src/ENDFtk/section/6/DiscreteTwoBodyScattering/test/DiscreteTwoBodyScattering.test.cpp
```diff
@@ -48,7 +48,7 @@ SCENARIO( "DiscreteTwoBodyScattering" ) {
 
         std::string buffer;
         auto output = std::back_inserter( buffer );
-        chunk.print( output, 9228, 6, 5 );
+        chunk.print( output, 9228, 26, 5 );
```
nathangibson14, line 51: You changed all the 6s to 26s.  I don't think you meant to do this.

nathangibson14 (changes requested):
Just a few pedantic comments.

src/ENDFtk/section/26/src/ctor.hpp
whaeck: The python bindings compile so it might not be required. I'll still add it to be safe.

nathangibson14 (approved):

PR #148, opened 2022-08-15
Feature/python extension

whaeck:
A small extension on the python bindings for use in resonance reconstruction


whaeck:
Setting this to draft since we don;t need this just yet

PR #152, opened 2022-09-20
Update/hana dep

whaeck:
Replaces the hana-adapter dependency with hana v1.7.0, the latest stable hana release.


cmake/develop_dependencies.cmake
```diff
@@ -23,9 +23,9 @@ FetchContent_Declare( disco
     GIT_SHALLOW     TRUE
     )
 
-FetchContent_Declare( hana-adapter
-    GIT_REPOSITORY  https://github.com/njoy/hana-adapter
-    GIT_TAG         origin/master
+FetchContent_Declare( hana
+    GIT_REPOSITORY  https://github.com/boostorg/hana
+    GIT_TAG         v1.7.0
```
nathangibson14, line 28: Why is this different from the release_dependencies file?

whaeck: You mean the tag value? I don't know, I used update_repository to generate the release file

whaeck: I wouldn't be surprised that the v1.7.0 commit also has that other boost tag (Boost 1.75) and your tool picked up on that one instead.

nathangibson14: Indeed, I just checked.  All these tags are the same:
tag: v1.7.0, tag: boost-1.76.0.beta1, tag: boost-1.76.0, tag: boost-1.75.0.beta1, tag: boost-1.75.0

So I guess this is fine.  Just making sure I can still compile, etc.

nathangibson14:
Just to capture the discussion we've had elsewhere, this is an opportunity to remove the compiler warnings that arise in GCC-12.  It appears that moving to an even newer hana will alleviate this, and the `boost-1.77.0` tag works.  Upgrading past that does not work out of the box, so let's see if we can figure that out quickly before pulling this in.

whaeck:
I agree. If we cannot figure out how to get the latest hana from boost 1.80 to interface properly, we can keep 1.77 for now.

whaeck:
Maybe another solution would be to continue to use the hana-adapter and update that one to the latest hana version. I do want to use the latest hana version to see if it resolves more hana compiler warnings in ENDFtk on my M1.

whaeck:
The hana dependency is getting removed

PR #153, opened 2022-10-11
Fix/update directory

whaeck:
Fixing an issue in the updateDirectory() function that caused the directory for MF1 to be omitted.


python/src/tree/Material.python.cpp
```diff
@@ -296,12 +296,15 @@ void wrapTreeMaterial( python::module& module, python::module& viewmodule ) {
   .def(
 
     "update_directory",
-    [] ( Material& self ) { return njoy::ENDFtk::tree::updateDirectory( self ); },
+    [] ( Material& self, bool copy_mod = false )
+       { njoy::ENDFtk::tree::updateDirectory( self, copy_mod ); },
```
ptalou, line 300: Why did you remove the "return" command?

ptalou:
Better description of what you are trying to do in this PR would be welcome.

whaeck:
> Better description of what you are trying to do in this PR would be welcome.

When using the update_directory function, MF1 entries (except MF1 MT451) were actually erased from the directory. This change (and the assocaited tests) ensures that this no longer happens.

ptalou (approved):

python/src/tree/Material.python.cpp
whaeck: The update_directory does not return anything, it is just a void function so the return statement does nothing.

PR #155, opened 2022-12-09
Feature/mf32 part2

whaeck:
Adding the CompactCorrelationMatrix, the class that hides the absurdity that are the INTG records to compact a correlation matrix. This class will ensure a user will NEVER see these abominations - unless they look in the ENDF file of course.

As an abstraction of the interface, CompactCorrelationMatrix will return the actual correlation coefficients and not the compacted integer values. You're welcome.


whaeck:
One year waiting for a review. I'm admin merging this.

PR #159, opened 2023-02-01
Feature/mf32 part5

whaeck:
This finishes the code for LCOMP = 2 but does not include full testing or the python bindings for the general and compact (LCOMP = 1 and 2) formats.

Part 6 will add the missing tests and python bindings.


whaeck:
One year waiting for a review. I'm admin merging this to get it over with.

PR #164, opened 2023-02-14
Feature/mf32 part7

whaeck:
Another part of the MF32 work.

Included here:
- Python bindings for most of the LCOMP=2 components
- Python tests and C++ unit tests
- Some corrections following the unit tests: the compact correlation matrix did not generate the right numbers for negative correlation values

There will be only one more PR after this one - I hope.


whaeck:
8 months waiting for a review. I'm admin merging this to get it over with.

PR #165, opened 2023-02-16
Update/cmake

whaeck:
Updating the CMake files to do the following:
- add an ENDFtk.python option to switch on or off the python API (default = ON). FYI: when including ENDFtk, setting this to OFF will remove it as a target
- renamed the unit test option to ENDFtk.tests (default = OFF)
- added the python unit tests to the ENDFtk.python target. SInce these tests are not compiled, switching them on when ENDFtk.python is compiled is not a waste of time
- -fPIC is now set using the proper CMake option (set_target_properties( ENDFtk.python PROPERTIES POSITION_INDEPENDENT_CODE ON))

As a result, the CI was updated again to just make and ctest instead of the older compile twice approach.


ptalou (approved):
Approved under the condition that every test passes (to be completed)

PR #166, opened 2023-02-22
Feature/mf32 part8

whaeck:
Part 8 of the MF32 work.

This mainly adds C++ tests for some components and python bindings for missing components (other than unresolved which will be in part 9 so I lied in the previous PR: there's going to be at least one more).


whaeck:
8 months waiting for a review. I'm admin merging this to get it over with.

PR #167, opened 2023-03-01
Feature/mf32 part9

whaeck:
The final part of the MF32 work:
- unresolved covariance data python bindings and tests
- LCOMP=1 tests

Some updates due to misunderstood format specs.

Correction to the ComapctCorrelationMatrix following tests by NNL (improper check for number of values to be put in an INTG record).


whaeck:
7 months waiting for a review. I'm admin merging this to get it over with.

PR #170, opened 2023-03-22
Update/mf35 rebase

whaeck:
updating the feature/mf35 branch using the latest branch that includes MF26, MF32, the update to the build system for python tests, etc.

The following changes were made to MF31/MF35/MF40:
- Common components were lifted out of MF33 (SquareMatrix, RectangularMatrix and CovariancePairs) in the MF32 work so the MF31/MF35/MF40 components were updated to reflect this change
- tests for these common components are removed
- python unit tests are now handled in a cmake include file

I made sure that all unit tests are included - both on the python and C++ side of things.

This PR contains ALL MF26 and MF32 work since it was based on the last branch. You only need to look at MF31/MF35/MF40 to understand these changes.

Once this is in the MF35 branch, I'll change the target branch for feature/mf35 from develop to feature/gehana. That way, we're back in order - and there will be only one with everything.


whaeck:
@nathangibson14 This is just the rebase of the mf35 branch using what was done in the latest mf32 pieces up to the removal of the hana dependency. There are indeed still common covariance pieces floating around in multiple sections at this point of the chain. I did not want to remove these other common pieces in this PR to make it easier to move forward. Removing the common covariance pieces is continued in the next PR (feature/covariance-commons).

PR #171, opened 2023-03-24
Feature/covariance commons

whaeck:
Moved some common covariance components from MF31/MF33/MF40 to the other common components.

This removes some Python binding code - which is good :-)


whaeck:
7 months waiting for a review. I'm admin merging this to get it over with.

PR #173, opened 2023-03-30
Feature/has section

whaeck:
This extends the tree and parsed material object interface by adding a has_mf_mt and has_section function that takes both the MF and MT number of a section.

Also adds a section(mf,mt) function that returns a Section object on the tree Material. Adding this to a parsed Material is slightly more complicated so I'll wait to add it.


whaeck:

PR #174, opened 2023-04-03
Adding the section(mf,mt) function for the parsed material object

whaeck:
This adds the section(mf,mt) function to the parsed material interface.


whaeck:

PR #176, opened 2023-05-29
Feature/python examples

whaeck:
Adding jupyter notebook examples, as used in the ENDFtk paper.

A few updates are made in the fission yield sections (updated comments, added human readable function that was missing).


whaeck:

PR #178, opened 2023-06-16
Feature/mf3 convenience

whaeck:
Adding a few convenience constructors for MF3, MF9 and MF10 where only an interpolation type is given instead of boundaries and interpolants.


whaeck:
5 months waiting for a review. I'm admin merging this to get it over with.

PR #181, opened 2023-10-17
Feature/cmake update

whaeck:
Updating the cmake files to be in line with how scion and ENDFtk are set up. Follow up PRs will do some work on the dependencies.


ptalou (approved):
Good to go!

PR #191, opened 2024-01-24
Feature/za

whaeck:
All ZA identifiers are now integers across all of ENDFtk:
- ZA in HEAD records,
- ZAI in isotopes
- ZAP in reaction products in MF6 and MF26

This will allow for easier searching on ZA values without having to do floatign point comparisons.

A human readable function alternative for ZA() was added: targetIdentifier(). Functions were added to MF6 and MF26 to search for specific reaction product identifiers (similar to the MF9 and MF10 setup).

Removed unused code and removed the setters on the basic records (HEAD, LIST, etc. are now fully immutable).


python/test/MF1/Test_ENDFtk_MF1_MT458_Section.py
```diff
@@ -180,8 +181,9 @@ def verify_chunk_polynomial( self, chunk ) :
             self.assertEqual( 458, chunk.MT )
             self.assertEqual( 458, chunk.section_number )
             self.assertEqual( 92235, chunk.ZA )
-            self.assertEqual( 233.0248, chunk.atomic_weight_ratio )
+            self.assertEqual( 92235, chunk.target_identifier )
             self.assertEqual( 233.0248, chunk.AWR )
+            self.assertEqual( 233.0248, chunk.atomic_weight_ratio )
```
ptalou, line 186: Why is this one not self.assertAlmostEqual for the atomic_weight_ratio (double)?

python/test/MF1/Test_ENDFtk_MF1_MT460_Section.py
```diff
@@ -116,8 +117,9 @@ def verify_chunk_lo2( self, chunk ) :
             self.assertEqual( 460, chunk.MT )
             self.assertEqual( 460, chunk.section_number )
             self.assertEqual( 92235, chunk.ZA )
-            self.assertEqual( 233.0248, chunk.atomic_weight_ratio )
+            self.assertEqual( 92235, chunk.target_identifier )
             self.assertEqual( 233.0248, chunk.AWR )
+            self.assertEqual( 233.0248, chunk.atomic_weight_ratio )
```
ptalou, line 122: same question

python/test/MF2/Test_ENDFtk_MF2_MT152_Section.py
```diff
@@ -33,6 +33,7 @@ def verify_chunk( self, chunk ) :
             self.assertEqual( 152, chunk.section_number )
 
             self.assertAlmostEqual( 94239., chunk.ZA )
+            self.assertAlmostEqual( 94239., chunk.target_identifier )
```
ptalou, line 36: Shouldn't those be self.assertEqual with 94239 as integer?

src/ENDFtk/section/15.hpp
ptalou, line 1: Is there actually a difference? Perhaps just the tabulation?

src/ENDFtk/section/2/151/Isotope/src/ctor.hpp
```diff
@@ -22,7 +22,7 @@ Isotope( ControlRecord&& cont,
          int MAT,
          int MF,
          int MT ) :
-  Isotope( cont.C1(), cont.C2(), cont.L2(),
+  Isotope( static_cast< int >( std::round( cont.C1() ) ), cont.C2(), cont.L2(),
```
ptalou, line 25: Hopefully round() does the job we talked about... int(x+0.5) should be identical

src/ENDFtk/section/5.hpp
ptalou, line 1: different tabulation?

src/ENDFtk/section/8/459/test/459.test.cpp
```diff
@@ -161,7 +161,7 @@ SCENARIO( "section::Type< 8, 459 >" ) {
 
     WHEN( "the data is given explicitly using arrays" ) {
 
-      double zaid = 92235.;
+      double zaid = 92235;
```
ptalou, line 164: Shouldn't this be an integer? Same question for the next two changes

ptalou (changes requested):
Please check my few comments and questions.

python/test/MF1/Test_ENDFtk_MF1_MT458_Section.py
whaeck: It should be, indeed.

src/ENDFtk/section/15.hpp
whaeck: Yes, this is just whitespace

python/test/MF1/Test_ENDFtk_MF1_MT458_Section.py
whaeck: with the exception of the targetIdentifier() function being added.

python/test/MF2/Test_ENDFtk_MF2_MT152_Section.py
whaeck: yup

src/ENDFtk/section/15.hpp
whaeck: with the exception of the targetIdentifier() function being added.

src/ENDFtk/section/5.hpp
whaeck: yes, but also the targetIdentifier() function being added.

src/ENDFtk/section/8/459/test/459.test.cpp
whaeck: yes, it should

ptalou (approved):

PR #194, opened 2024-03-12
Fix unit test still using catch-adapter

whaeck:
A unit test (WattSpectrum.test.cpp in MF5) was still using the old Catch version. On gcc 11.4, this seemed to cause a compilation issue. This has been corrected.

In addition, a release notes file was added to more easily keep track of changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,12 +47,17 @@ The CMake-based build system will try to automatically detect the installed vers
 
 A version of python 3.x is preferred.
 
-##### importError cannot import name <sysconfig> #####
+##### ImportError cannot import name <sysconfig> #####
 
 This error sometimes comes up when running the cmake command. This appears to be related to an incomplete/corrupted python installation. It can be rectified by installing the distutils package for the python version that is being used. On a linux system, the following command should install the distutils package:
 ```
 sudo apt install python3-distutils
 ```
+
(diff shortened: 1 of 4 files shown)


README.md
Camden-Blake, line 1: A problem may occur if a user's installed python environment is through conda.  The user will be able to build properly but all of the unit tests for python may fail.  The error that I experienced was that ENDFtk could not be imported due to GLIBCXX_3.4.30 missing.  This seems to be an a problem with conda not having GLIBCXX_3.4.30 in the path.  If the user is able to compile ENDFtk, they have GLIBCXX_3.4.30.  At which point, a symlink will need to be created to point the conda enviroment to where GLIBCXX_3.4.30 is installed.

This is what fixed it for me: https://askubuntu.com/questions/1418016/glibcxx-3-4-30-not-found-in-conda-environment

Camden-Blake (commented):
A problem with conda may stop the import of ENDFtk into a python script.  A symlink needs to be created to allow the import.  Not a problem with this implementation but might be note worthy for the README.md

README.md
whaeck: I added some troubleshooting info in the readme for this. I'm hesitant on adding the link though for security reasons.

Camden-Blake (approved):

Issue #202, opened 2024-06-12
MF8 MT457 NT validation


thomasms:
Hi Wim,

I am trying to use ENDFtk (from develop branch - commit 936a5861fd04f8a7179a59f0f0c3a2091482a3b0) to parse JEFF33 decay data from [here](https://www.oecd-nea.org/dbdata/jeff/jeff33/index.html#decay_data).

Running the following script I get an error for MAT 128.
![image](https://github.com/njoy/ENDFtk/assets/8820266/b4087d89-6ce4-40f0-be30-672a06fd3695)

```python
import argparse
import ENDFtk as tk


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("decfile")
    args = parser.parse_args()

    tape = tk.tree.Tape.from_file(args.decfile)
    print(f"Found {len(tape.material_numbers)} materials")
    print("Attempting to parse each material...")
    for mat in tape.materials:
        try:
            pmat = mat.parse()
        except Exception:
            print(f"Could not parse {mat.material_number}")


if __name__ == '__main__':
    main()
```

In this case NT is equal to 8 and according to the implementation [here](https://github.com/njoy/ENDFtk/blob/develop/src/ENDFtk/section/8/457/DiscreteSpectrum/src/verifyNT.hpp) only NT values of 4, 6, and 12 are permitted.

Thanks to a very knowledgeable nuclear data colleague, he noted that NT=8 can be permitted in some cases according to the ENDF manual section 8.4.2 point 5.a: _"The amount of additional information depends upon the detail in which quantities were obtained for inclusion in STYP=8. or 9. spectra, and the number of decay modes. (This detail will also be reflected in the uncertainties assigned in STYP=8. or 9. spectra.) If only the total conversion electron emission is calculated, RICC and RICC should be included and NT is specified as 8."_

To me this suggests that this could be a bug in ENDFtk, is there any way to include this NT=8 corner case in ENDFtk?

thomasms:
Whoops duplicate of https://github.com/njoy/ENDFtk/issues/175

whaeck:
Yes, this is correct. NT can be equal to 8, as I discovered some time ago myself: #175. I've been putting off correcting it, so I'll try to do it today.

As a side note: we will be doing another release soon with a few corrections so I'll try to include this in there as well.

PS: on the GitHub emails, you come up as Tom and I find that confusing as that's my brother's name ;-)

whaeck:
[...]

PR #203, opened 2024-06-12
Adding the possibility of using NT=8 in MF8 MT457 DiscreteSpectrum

### LINKED ISSUE #175: NT values for MF8 MT457 discrete spectra (decay data)
The procedures part of the ENDF manual MF8 MT457 states that NT can be 8. In the latest version/branch of ENDFtk, we verify if NT is equal to 4, 6 or 12.

We should verify and change the test function and the appropriate constructors.

whaeck:
This should fix #175 and #202 

@thomasms

### DIFF
--- a/.github/workflows/ContinuousIntegration.yml
+++ b/.github/workflows/ContinuousIntegration.yml
@@ -16,8 +16,8 @@ jobs:
     runs-on: ${{matrix.os}}
     strategy:
       matrix:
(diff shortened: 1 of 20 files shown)


thomasms:
That was fast - it seems you can implement a fix faster than it takes me to compile ENDFtk 😀

Magic, I can now parse the JEFF 33 file without issue! Thanks

PR #208, opened 2024-07-03
Feature/map to list

whaeck:
Another PR to prepare for the removal of range-v3.

The tree::Tape, tree::File, tree::Section and the parsed equivalents all use std::map or std::multimap to store their information and we use views::keys and views::elements to access the data as ranges (bidirectional ones).

Trials with the replacement of range-v3 have shown issues with the views::keys and views::elements replacements that cause compilation errors. Since these are basically implementation details, I decided to replace the maps and multimaps with lists. This does not break the interface as the entire interface still behaves in the same way (we can iterate over materials, files, sections, etc. as bidirectional ranges).

I did try using vector instead of list but the removal and insertion of sections/files/materials was broken by this change. Lists allow elements to be inserted without invalidating iterators while vectors do not. As a result, keeping a reference around to an element after inserting an element in a vector may be dangerous. Because of this, I went back to lists which maintains the current interface.

### DIFF
--- a/src/ENDFtk/file/Base.hpp
+++ b/src/ENDFtk/file/Base.hpp
@@ -2,10 +2,11 @@
 #define NJOY_ENDFTK_FILE_BASE
 
 // system includes
-#include <map>
+#include <list>
 
 // other includes
-#include "range/v3/view/map.hpp"
+#include "range/v3/view/all.hpp"
+#include "range/v3/view/transform.hpp"
 
 namespace njoy {
 namespace ENDFtk {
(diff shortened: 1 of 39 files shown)


src/ENDFtk/file/Base.hpp
```diff
@@ -118,18 +99,17 @@ namespace file {
...
-                    " correspond to a stored section" );
-        Log::info( "Requested section number: {}", mt );
-        Log::info( "File queried: ", this->MF() );
-        throw o;
-      }
+      Log::error( "Requested section number (MT) does not"
+                  " correspond to a stored section" );
+      Log::info( "Requested section number: {}", mt );
+      Log::info( "File queried: ", this->MF() );
+      throw std::exception();
```
cjosey, line 112: Do you think it's worthwhile to add information to these (and other related new bare throws), or just rely on the `Log` calls above?

cjosey (approved):
Overall, looks good. Ran over it with UBSan to make sure iterator issues are defined, and they were.

src/ENDFtk/file/Base.hpp
whaeck: This is how it was done in the past, and in every NJOY library. I'm not a fan of it either and it is something that I want to revisit at some point.

PR #215, opened 2024-09-26
Vixit - removing range-v3 as a dependency

whaeck:
This PR removes range-v3 as a dependency. It is replaced with the partial std20 and std23 implementation for the ranges standard. Only a very few ENDF components had to be revised (read: hacked) to fix some Python issues or issues related to range-v3 components that were not in the standard. This concerns the following:
- polynomial energy release components
- some components in MF7
- MF2 MT152 (an internal NJOY format)
- some MF6 LAW=1 components
- some MF4 components

Other than these last components, most of the changes are very straight forward (essentially small interface changes). Once we can move to C++20/C++23, the update will be trivial as the tools implementation uses the C++20/C++23 interfaces.


src/ENDFtk/section/1/458/PolynomialComponents/src/print.hpp
```diff
@@ -0,0 +1,26 @@
...
+ *
+ *  @tparam OutputIterator   an output iterator
+ *
+ *  @param[in] it            the current position in the output
+ *  @param[in] MAT           the MAT number
+ *  @param[in] MF            the MF number
+ *  @param[in] MT            the MT number
+ */
+template< typename OutputIterator >
+void print( OutputIterator& it, int MAT, int MF, int MT ) const {
```
joewpeterson, line 12: A relatively easy way to improve performance would be to reserve size the vector to (this->order() + 1) * 9 *2. This should save lg((this->order() + 1) * 9 *2) heap allocations.

While I'm thinking of it, I think the best practices for added to vectors now is to use emplace_back. It shouldn't make any difference for this case because you are pushing doubles, but I think the modern c++ wisdom is to abandon push_back in favor of emplace_back in all cases as the worst case scenario is equivalent performance. It's something I have to be reminded of every now and again

whaeck: Yes, I do use emplace_back for anything other than primitive types (at least I try).

I added the reserve() as well, although for smaller vectors like this one it probably does not matter (the polynomial order is often just 0 or 1 in this case) - as vectors grow in blocks (there is a default capacity being reserved which gets added to whenever one goes over capacity). The size of that block does depend on implementation - I think.

joewpeterson: Sounds good. Yeah, I don't know how often this is expected to be called but anything we do to reduce number of heap allocs will increase speed of each call. 

Looks like there are other places where the same comment can be made.  For example, in generateList function here does the same thing and could benefit from a reservation. Might be worth opening a task to look for cases of this as there might not be instances of it I'm seeing in changed files presented here? 

This is the only improvement I've spotted and and this does pass the memory leaks test. So if an optimization task could be created to address these vector reservations, I'll approve this one

whaeck: The print() function might not get called too often - only when we insert it into a tape, or if we print it to a file.

Yes, we can take a note of the reserve(). It's a lot of code to go over though ;-)

I'm still working on MF4 - I can't seem to get that one to work for some reason. You should make sure that you have checked "viewed" on all files before I push those changes.

python/src/section/4.python.cpp
```diff
@@ -23,8 +23,8 @@ void wrapSection_4( python::module& module, python::module& viewmodule ) {
   // type aliases
   using Section = njoy::ENDFtk::section::Type< 4 >;
   using Distributions = Section::Distributions;
-  using Distribution = Section::Variant;
-  using DistributionRange = BasicRandomAccessAnyView< Distribution >;
+//  using Distribution = Section::Variant;
```
joewpeterson, line 26: Perhaps for commented out code it would be better to either delete it or include a comment on why this is commented out and left in? 

This is seen in a number of places

whaeck: These are the pieces I am still working on. I'll be done with this in a couple of hours. There should be no commented out code left.

src/ENDFtk/section/1/458/PolynomialComponents/src/ctor.hpp
```diff
@@ -1,18 +1,20 @@
...
+PolynomialComponents( int nply, std::vector< double >&& list ) :
+  nply_( nply ), values_( std::move( list ) ) {}
 
 /**
  *  @brief Private constructor
  */
 PolynomialComponents( std::vector< double >&& list ) :
-  ListRecord( 0.0, 0.0, 0, list.size() / 18 - 1, list.size() / 2,
-              std::move( list ) ) {}
+  PolynomialComponents( list.size() / 18 - 1, std::move( list ) ) {}
```
joewpeterson, line 11: I think use of magic number 18 corresponds to PolynomialComponents, yes? Is there any chance this will ever change? If yes, we definitely would prefer some definition of this size we place in a common location. Might be better programming to do that anyway

src/ENDFtk/section/1/458/PolynomialComponents/src/generateList.hpp
```diff
@@ -32,26 +32,50 @@ generateList( std::vector< std::array< double, 2 > >&& efr,
...
 }
+
+static std::vector< double >
+generateList( const ListRecord& list ) {
+
+  std::vector< double > values;
+
+  int npl = list.NPL();
+  int nply =  list.L2();
+  if ( npl != ( nply + 1 ) * 18 ) {
```
joewpeterson, line 64: While I'm thinking about the magic number issue, it might be good to at least have comment on why 9 is chosen.

src/ENDFtk/section/1/458/PolynomialComponents/src/ctor.hpp
whaeck: I like how you thing about magic numbers, I'm not a fan either :-)

This will/should never change - although it would just take one evaluator to get it changed. There are 9 components to the fission Q values - as defined in the ENDF manual (see https://www.nndc.bnl.gov/endfdocs/ENDF-102-2023.pdf). The magic number of 18 is the number of these components times 2, for one value and one uncertainty for each component.

src/ENDFtk/section/2/151/RMatrixLimited/ParticlePairs.hpp
```diff
@@ -16,9 +16,10 @@ class ENDFTK_PYTHON_EXPORT ParticlePairs : protected ListRecord {
 
   auto column( unsigned int i ) const {
 
+    using namespace njoy::tools;
     return ListRecord::list()
-             | ranges::views::drop_exactly( i )
-             | ranges::views::stride( 12 );
+             | std20::views::drop( i )
+             | std23::views::stride( 12 );
```
joewpeterson, line 22: Not certain why I wasn't thinking of magic numbers before, but now I am. It would probably be best if why these numbers are chose could be self evident in code. Perhaps an enum list could be used for this section as a means to document why each value is chosen?

src/ENDFtk/section/2/151/RMatrixLimited/ResonanceChannels.hpp
```diff
@@ -17,9 +17,10 @@ class ENDFTK_PYTHON_EXPORT ResonanceChannels : protected ListRecord {
 
   auto column( unsigned int i ) const {
 
+    using namespace njoy::tools;
     return ListRecord::list()
-             | ranges::views::drop_exactly( i )
-             | ranges::views::stride( 6 );
+             | std20::views::drop( i )
+             | std23::views::stride( 6 );
```
joewpeterson, line 23: Similar comment here on the number. I don't think this issue should not prevent the merger for this of this branch, as the branch is about replacing v3. But perhaps we can create new tasks to address these?

src/ENDFtk/section/2/151/RMatrixLimited/TabulatedBackgroundRMatrix.hpp
```diff
@@ -109,11 +117,14 @@ class ENDFTK_PYTHON_EXPORT TabulatedBackgroundRMatrix :
    */
   auto RB() const {
 
-    return ranges::views::zip_with(
-               [] ( double real, double imag ) -> std::complex< double >
-                  { return { real, imag }; },
-               this->real_.y(),
-               this->imaginary_.y() );
+    // using namespace njoy::tools;
```
joewpeterson, line 120: Commented out code should be removed or should be commented on as to why it is commented out

src/ENDFtk/section/2/151/UnresolvedEnergyDependent/JValue.hpp
```diff
@@ -14,10 +14,11 @@ class ENDFTK_PYTHON_EXPORT JValue : protected ListRecord {
 
   auto column( const int index ) const {
 
+    using namespace njoy::tools;
     return
       ListRecord::list()
-        | ranges::views::drop_exactly( index + 6 )
-        | ranges::views::stride(6);
+        | std20::views::drop( index + 6 )
```
joewpeterson, line 20: Include into task to document and/or explain why 6 is used

src/ENDFtk/section/2/151/UnresolvedEnergyDependentFissionWidths/JValue.hpp
```diff
@@ -127,7 +127,8 @@ class ENDFTK_PYTHON_EXPORT JValue : protected ListRecord {
    */
   auto GF() const {
 
-    return ListRecord::list() | ranges::views::drop_exactly( 6 );
+    using namespace njoy::tools;
+    return ListRecord::list() | std20::views::drop( 6 );
```
joewpeterson, line 131: Include into task to document and/or explain why 6 is used

src/ENDFtk/section/2/152.hpp
```diff
@@ -133,9 +139,10 @@ namespace section {
      */
     auto EUNR() const {
 
+      using namespace njoy::tools;
       return this->data_.list()
-               | ranges::views::drop_exactly( this->NSIGZ() )
-               | ranges::views::stride( 1 + this->NSIGZ() * 5 );
+               | std20::views::drop( this->NSIGZ() )
+               | std23::views::stride( 1 + this->NSIGZ() * 5 );
```
joewpeterson, line 145: Include into task to document and/or explain why 5 is used

src/ENDFtk/section/28/SubshellData.hpp
```diff
@@ -12,8 +12,9 @@ class ENDFTK_PYTHON_EXPORT SubshellData : protected ListRecord {
 
   auto column( unsigned int i ) const {
 
-    return ListRecord::list() | ranges::views::drop_exactly( 6 + i )
-                              | ranges::views::stride( 6 );
+    using namespace njoy::tools;
+    return ListRecord::list() | std20::views::drop( 6 + i )
+                              | std23::views::stride( 6 );
```
joewpeterson, line 17: Include into task to document and/or explain why 6 is used

src/ENDFtk/section/32/151/CompactBreitWignerUncertainties.hpp
```diff
@@ -14,8 +14,9 @@ class ENDFTK_PYTHON_EXPORT CompactBreitWignerUncertainties :
 
   auto column( unsigned int i ) const {
 
-    return ListRecord::list() | ranges::views::drop_exactly( i )
-                              | ranges::views::stride( 12 );
+    using namespace njoy::tools;
+    return ListRecord::list() | std20::views::drop( i )
+                              | std23::views::stride( 12 );
```
joewpeterson, line 19: Include into task to document and/or explain why 12 is used

src/ENDFtk/section/4.hpp
```diff
@@ -62,18 +58,18 @@ namespace section {
                             TabulatedDistributions,              // LTT=2, LI=0
                             MixedDistributions >;                // LTT=3, LI=0
 
-    using Variant =
-      std::variant< std::reference_wrapper< const LegendreCoefficients >,
-                    std::reference_wrapper< const TabulatedDistribution > >;
+//    using Variant =
```
joewpeterson: Commented out sections could be deleted

src/ENDFtk/section/4/MixedDistributions.hpp
```diff
@@ -83,55 +83,55 @@ class ENDFTK_PYTHON_EXPORT MixedDistributions {
...
-   *  @brief Return the incident energy values
-   */
-  auto incidentEnergies() const {
-
-    return ranges::views::concat( this->legendre().incidentEnergies(),
-                                  this->tabulated().incidentEnergies() );
-  }
-
-  #include "ENDFtk/section/4/MixedDistributions/src/angularDistributions.hpp"
+//  /**
```
joewpeterson: Commented out section could be deleted

whaeck: That's the MF4 coding I just corrected, you may want to refresh your browser.

src/ENDFtk/section/7/2/CoherentElastic/src/generateSandT.hpp
```diff
@@ -0,0 +1,12 @@
+void  generateSandT() {
+
+  this->s_.clear();
+  this->t_.clear();
+  this->s_.push_back( this->principal_.y() );
+  this->t_.push_back( this->principal_.C1() );
```
joewpeterson, line 6: Another location for performance gain? Perhaps the number of allocations will be fewer becasue clear() will keep the reservation, but emplace_back should be used for the new population for at least s_ I think. These are type principal_.y() which are being copied, yes? We'd prefer them to be moved?

src/ENDFtk/section/7/4/TabulatedFunctions/ScatteringFunction/src/generateSandT.hpp
```diff
@@ -0,0 +1,12 @@
+void  generateSandT() {
+
+  this->s_.clear();
+  this->t_.clear();
+  this->s_.push_back( this->alphas_.y() );
+  this->t_.push_back( this->alphas_.C1() );
```
joewpeterson, line 6: More performance gain possible? depending on details. clear with keep prior allocation. Is it guaranteed to be the right size? structure of s_ may prefer emplace_back

src/ENDFtk/section/2/151/RMatrixLimited/TabulatedBackgroundRMatrix.hpp
whaeck: The note above explains this, but I'll add another line

joewpeterson (approved):
latest rendition has been tested for memory leaks locally and passes the test

PR #217, opened 2024-12-12
Initial Windows Support for ENDFtk

HunterBelanger:
This PR makes a few modifications so that ENDFtk can be compiled natively on Windows. For now, the changes have been quite minor:
* Use development branch of NJOY tools to avoid problems with auto returns in the logger
* Change two methods inside `InterpolationBase` to avoid MSVC compiler errors due to calling previously defined auto return methods
* Change a CMake setting to avoid MSVC complaining about unknown compiler flags
* Update two unit tests that were failing due to iterator invalidation

After making these changes, all the tests passed on my system (Windows 11 with MSVC 19.40). There are still, however, a few issues to address:
* Windows and CMake were constantly complaining about file paths which are too long (almost exclusively in the unit tests). Because of this, some of the tests just wouldn't even be built. To fix this, I had to update my Windows registry to allow longer file names. This allowed those tests to be built, and the passed, but CMake and Windows is still complaining about paths that are too long and can't be tracked.
* Initially, two unit tests were failing. After examination, it appeared that this was likely due to iterator invalidation when adding strings together in the test. Fixing this which allowed the tests to pass. I have not checked to see if this type of problem exists in other tests though, so this might be something that needs to be addressed elsewhere.
* For now, I have just set the dependencies to pull the development branch of NJOY tools. I made a parallel PR there to fix the unknown compiler flag warnings in that package. At some point, we will need to set this dependency to a commit hash.

@whaeck I am now sure how you want to address these remaining problems (primarily the first one) ?


whaeck:
@HunterBelanger We're working on this internally and setting up a Windows environment to be able to compile on Windows with MSVC so that we can do some checks.

whaeck:
At the very least we want to get MSVC to be up and running on Windows. We can look into the intel issue once that is done.

We have an internal branch with most of the changes of this PR in it (as I mentioned above, we're left with the InterpolationBase auto issue). Once these changes go in develop internally, we can close this PR.

whaeck:
@HunterBelanger We've just pushed changes to develop that take into account all these issues. Internally, we're seeing more problems when using MSVC but I wanted to get these changes out so that you could test them. Let me know if everything compiles for you, and if the unit tests all succeed.

HunterBelanger:
I will definitely try out those changes and see what they give for me ! I won't be able to take a look at it until the first week of June, however, with some upcoming travel.

whaeck:
I hope we'll be able to meet when you're at LANL next week.

whaeck:
All changes are now in main.

